PR #225, opened 2023-07-12
Test musica

K20shores:
Builds CAM-SIMA with musica. If you look at the output of the cesm log and search of "new state", you'll see the result of print statements we have from the micm component. Again, not much happens as of now with the component, but the connections are there


nusbaume:
@K20shores I think I know what is going on with the failed test.  Would it be alright if pushed the fix to your branch?

cacraigucar (commented):
My limited 2 cents worth review

Externals_CAM.cfg
```diff
@@ -22,8 +22,8 @@ required = False
 [ncar-physics]
 local_path = src/physics/ncar_ccpp
 protocol = git
-repo_url = https://github.com/NCAR/atmospheric_physics
-tag = atmos_phys0_00_018
+repo_url = https://github.com/K20shores/atmospheric_physics
+branch = musica_ccpp
```
cacraigucar, line 26: Typically we do not allow personal branches into CAM and I'm guessing the same holds for CAM-SIMA.  Of course y'all may be on top of this already

cacraigucar (commented):
Submitting this time with my comment

Externals_CAM.cfg
K20shores: I'm not sure what's proper protocol here. Musica is not yet merged into the atmospheric physics repo. Once it is, we can change the URL here

K20shores: That PR is [here](https://github.com/NCAR/atmospheric_physics/pull/63), just to establish a link in github

nusbaume: I agree that we can resolve this concern by simply getting the atmospheric_physics PR merged and tagged first before merging in these changes.  

Also, while it would be good to have the new MICM variables in the standard names dictionary soon, we can always get everything else merged in first and then always go back and update the names later in a separate PR if NOAA requests any changes.

nusbaume: @K20shores A new tag now exists for the `atmospheric_physics` mods you added, so I believe you can replace the current code changes with:

```suggestion
repo_url = https://github.com/NCAR/atmospheric_physics
tag = atmos_phys0_00_019
```

Which should resolve @cacraigucar's concern here.

cacraigucar:
I see that there is a netCDF file being included with this PR.   Typically we have prohibited files from being committed to the code repos and instead, files reside in inputdata.  Can the testing files reside in inputdata and be downloaded in the typical cime way?

K20shores:
@nusbaume Now that the other PR is merged, would you like me to copy this dockerfile into the `docker` folder and leave the dockerfile at the top level directory as one that builds the held suarez scheme?

K20shores:
@nusbaume Done. I moved the dockerfiles into the docker folder and updated instructions accordingly

cacraigucar:
Just an FYI to folks  (I'm not sure where the proper place is to document this).  During the manage_externals/checkout_externals step, I got the following message:

 => ERROR [ 9/21] RUN ./manage_externals/checkout_externals                                                        0.5s
 > [ 9/21] RUN ./manage_externals/checkout_externals:
0.483 /usr/bin/env: 'python\r': No such file or directory
0.483 /usr/bin/env: use -[v]S to pass options in shebang lines

@mwaxmonsky helped me and he finally figured out that my TortoiseGit (which I used to do my git clone/checkout on my Windows box) had an option to automatically convert line feeds into Windows line feeds.  I believe this is the default setting for TortoiseGit (but my admins installed this years ago, so I'm not completely sure).  Turning off the auto convert, allowed me to proceed with the manage_externals/checkout_externals step.

cacraigucar (approved):
From the usability standpoint, I approve this PR

nusbaume:
@K20shores I have some changes to `config_component.xml` that I want to make for this PR, but I figured it would be easier for me to just push them directly to the branch as opposed to making a bunch of change requests.  Is that OK with you, or would you rather I just add the requests to a review on Github instead?

.github/scripts/pr_mod_file_tests.py
```diff
@@ -59,28 +59,45 @@ def _file_is_python(filename):
...
+                                    #If the word exists, then assume
+                                    #it is a python file:
+                                    is_python = True
+                                #End if
+                            #End if
+                        #End if
 
                         #Exit loop, as only the first non-blank
                         #line should be examined:
                         break
```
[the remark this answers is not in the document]
nusbaume: Yes, good catch!  That was my mistake, so I can add the fix to my commit if Kyle is ok with it.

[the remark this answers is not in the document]
mwaxmonsky: Thanks!  I also forgot to grab the `#End if` above this block in this comment so the one above this block would have to be moved below it and stay at it's current indent level.

.github/scripts/pr_mod_file_tests.py
```diff
@@ -59,28 +59,45 @@ def _file_is_python(filename):
...
+
+                        #Ignore blank lines:
+                        if line.strip():
+
+                            #Check that first non-blank
+                            #line is a shabang:
+                            if line[0:2] == '#!':
+                                #If so, then check that the word
+                                #"python" is also present:
+                                if line.find("python") != -1:
```
mwaxmonsky, line 77: Would it be worthwhile to update this to the [recommended](https://peps.python.org/pep-0394/#for-python-runtime-distributors) pep8 specs to check for `python3`?

The following from the `pep8` doc might be the relevant reason if we are only supporting python3.7+:
> Some Linux distributions will not provide a `python` command at all by default, but will provide a `python3` command by default.

nusbaume: I think in this case I would rather keep it as just `python`, because that way this script will catch all python files that are being committed (given that `python` as a string will match all the different variants).  This will ensure that even if by accident we commit a python2 file it will still be linted.

However, I agree that for our own python scripts we should make sure we are using `python3` instead of just `python` when using shebang calls.  I'll go ahead and modify this script's shebang as a start.

docker/Dockerfile.musica
```diff
@@ -0,0 +1,119 @@
...
+###################################################
+## Install necessary packages
+###################################################
+RUN dnf -y update \
+    && dnf -y install \
+      ftp \
+      git \
+      hostname \
+      m4 \
+      python \
```
mwaxmonsky, line 14: Similar to the python comment above, would it be worthwhile to update this to `python3` to be explicit?  I think most linux distros just alias this at install time to check for `python3` so it might not be necessary.

.github/scripts/pr_mod_file_tests.py
mwaxmonsky, line 77: Probably a little more pedantic but it might be worth modifying these to be more pythonic unless this is hot code and needs as much performance as possible.

```suggestion
                            if line.startswith("#!"):
                                #If so, then check that the word
                                #"python" is also present:
                                if "python" in line:
```

nusbaume: Agreed.  I can bring this change in with my commits as well if @K20shores is OK with that.

K20shores:
> @K20shores I have some changes to `config_component.xml` that I want to make for this PR, but I figured it would be easier for me to just push them directly to the branch as opposed to making a bunch of change requests. Is that OK with you, or would you rather I just add the requests to a review on Github instead?

yes, please push directly to this branch!

docker/Dockerfile.musica
K20shores: This does install python3 by default. I'm good to be explicit if you prefer though

nusbaume (approved):
I've pushed the changes I wanted, so this PR is good on my end!

I should note that I tested my changes by building/running both the `cam-sima` and `musica` containers.  Still, it would be good to have others  re-build/run the containers themselves at least one more time to make sure I didn't break anything.  Thanks!

.github/scripts/pr_mod_file_tests.py
nusbaume: @mwaxmonsky this should be fixed now, but please let me know if that is not the case.

nusbaume: @mwaxmonsky The three Github-related scripts should now be using `python3` instead of `python`.

.github/scripts/pr_mod_file_tests.py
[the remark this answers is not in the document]
nusbaume: @mwaxmonsky this should be fixed now, but please let me know if that is not the case.

docker/Dockerfile.musica
mwaxmonsky: If it's not too much trouble, I'd be happy to push the change directly to the branch.

K20shores: Fine with me!

mwaxmonsky: So when changing this to `python3`, the build produces the log output: `Package python3-3.11.4-1.fc38.x86_64 is already installed.`  I originally thought this was fine but the build ultimately failed with:
```
STEP 18/37: RUN ./manage_externals/checkout_externals
/usr/bin/env: 'python': No such file or directory
Error: building at STEP "RUN ./manage_externals/checkout_externals": while running runtime: exit status 127
```

Leaving this as `python` is actually installing the package `python-unversioned-command`:
```
STEP 2/37: RUN dnf -y update     && dnf -y install       ftp       git       hostname       m4       python       sudo       svn       tree       vim     && dnf clean all
Fedora 38 - x86_64                               32 MB/s |  83 MB     00:02
Fedora 38 openh264 (From Cisco) - x86_64        2.6 kB/s | 2.5 kB     00:00
Fedora Modular 38 - x86_64                      935 kB/s | 2.8 MB     00:03
Fedora 38 - x86_64 - Updates                     12 MB/s |  30 MB     00:02
Fedora Modular 38 - x86_64 - Updates            2.0 MB/s | 2.1 MB     00:01
Dependencies resolved.
Nothing to do.
Complete!
Last metadata expiration check: 0:00:02 ago on Wed Aug 16 15:16:56 2023.
Package sudo-1.9.13-2.p2.fc38.x86_64 is already installed.
Dependencies resolved.
================================================================================
 Package                     Arch    Version                     Repo      Size
================================================================================
Installing:
 ftp                         x86_64  0.17-91.fc38                fedora    62 k
 git                         x86_64  2.41.0-1.fc38               updates   54 k
 hostname                    x86_64  3.23-8.fc38                 fedora    27 k
 m4                          x86_64  1.4.19-5.fc38               fedora   303 k
 python-unversioned-command  noarch  3.11.4-1.fc38               updates   11 k
 ...
```
which is what causes the build to succeed.
 
Given that this requires a change to the `checkout_externals` script, I'm okay leaving this as is instead.

mwaxmonsky (approved):
Looks good to me!  Apologies for the delay!

boulderdaze (approved):

Issue #208, opened 2023-02-13
Eliminate requirement of always having specific humidity


nusbaume:
### What is the feature/what would you like to discuss?

There is a requirement in CAM, and for the moment CAMDEN, that specific humidity always be present as an advected constituent.  However, for certain scenarios (such as a dry physics suite or an exoplanet simulation) there is no reason to have specific humidity, and thus requiring the tracer increases the computational cost and complexity of those simulations.

Thus it would be beneficial, and improve model flexibility, if specific humidity was treated like all other advected constituents, and only included when a physics scheme required it.

### Will this change (regression test) answers?

I Don't Know

### Will you be implementing this enhancement yourself?

Any Software Engineer can do this

PR #232, opened 2023-10-13
Update CAM-SIMA to run Kessler and Held-Suarez bit-for-bit with CAM

### LINKED ISSUE #210: Required variable standard name updates
### What happened?

After going through the CAM names with scientists, the following standard names need to be updated:

1. `reference_pressure_at_surface`             -> `reference_pressure`
2. `reference_pressure`                                 -> `reference_pressure_of_atmosphere_layer` (need to check with NOAA first)
3. `reference_pressure_normalized_by_surface_pressure` -> `reference_pressure_of_atmosphere_layer_normalized_by_surface_pressure` (need to check with NOAA first)
4. `rain_water_mixing_ratio_wrt_moist_air` -> `rain_mixing_ratio_wrt_moist_air`
5. `rain_water_mixing_ratio_wrt_dry_air`    ->  `rain_mixing_ratio_wrt_dry_air`
6. `snow_water_mixing_ratio_wrt_moist_air`  ->  `snow_mixing_ratio_wrt_moist_air`
7. `ln_of_air_pressure`                                  -> `ln_air_pressure`
8. `ln_of_air_pressure_at_interface`             -> `ln_air_pressure_at_interface`
9. `ln_of_air_pressure_of_dry_air`               -> `ln_air_pressure_of_dry_air`
10. `ln_of_air_pressure_of_dry_air_at_interface` -> `ln_air_pressure_of_dry_air_at_interface` 
11. `pressure_thickness`                                -> `air_pressure_thickness`
12. `pressure_thickness_of_dry_air`             -> `air_pressure_thickness_of_dry_air`
13. `geopotential_at_surface`                       -> `surface_geopotential`
14. `reciprocal_of_pressure_thickness`       -> `reciprocal_of_air_pressure_thickness`
15. `reciprocal_of_pressure_thickness_of_dry_air` -> `reciprocal_of_air_pressure_thickness_of_dry_air`
16. `x_wind`                                                     -> `eastward_wind`
17. `y_wind`                                                     -> `northward_wind`
18. `geopotential_height`                              -> `geopotential_height_wrt_surface`
19. `geopotential_height_at_interface`        -> `geopotential_height_wrt_surface_at_interface`
20. `tendency_of_x_wind_due_to_model_physics` -> `tendency_of_eastward_wind_due_to_model_physics`
21. `tendency_of_y_wind_due_to_model_physics` -> `tendency_of_northward_wind_due_to_model_physics`
22. `column_integrated_total_kinetic_and_static_energy_of_current_state` -> `vertically_integrated_energies_of_current_state_in_cam`
23. `column_integrated_total_kinetic_and_static_energy_of_initial_state` -> `vertically_integrated_energies_of_initial_state_in_cam`
24. `column_integrated_total_water_of_current_state` -> `vertically_integrated_total_water_of_current_state`
25. `column_integrated_total_water_of_initial_state` -> `vertically_integrated_total_water_of_initial_state`

Along with these changes, it was noticed that in `src/dynamics/utils/hycoef.F90` that the `psr` and `ps0` variables are always set to match each other, and thus there is no reason to maintain both quantities.

Start talking to scientists

### What CAMDEN hash were you using?

101fd31

### What machine were you running CAMDEN on?

Other (please explain below)

Other (please specify below)

_No response_

Yes

### Extra info

This is a "bug" from a science standpoint, not a software standpoint.  Hence there isn't any particular run that this bug is associated with.

nusbaume:
This PR introduces a large number of standard name and other changes to allow one to run the new Kessler and Held-Suarez physics suites with CAM snapshots, and to ensure that they are bit-for-bit with CAM's results.

This PR also ensures that the SE dycore still runs with both Held-Suarez and Kessler physics suites (but it's values can't be formally checked yet due to the dycore being out-of-sync with CAM).

This PR also performs some code clean-up, and adds the ability for `check_field` to print the location (as an MPI rank, array column, and vertical level) of the maximum difference for each variable, which should help with debugging.

closes #210

## Tests run

Ran single timestep `FPHYStest` runs with CAM snapshots for Held-Suarez and Kessler with GNU on Derecho with the ne3pg3 grid, and got bit-for-bit results.  Also ran `FHS94` and `FKESSLER` compsets with the `ne5pg2` grid and the GNU compiler for five timesteps each, and both ran successfully.

### DIFF
--- a/cime_config/cam_autogen.py
+++ b/cime_config/cam_autogen.py
@@ -437,8 +437,8 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
         os.makedirs(physics_blddir)
     # End if
     # Collect all source directories
-    source_search = [source_mods_dir,
-                     os.path.join(atm_root, "src", "physics", "ncar_ccpp")]
+    atm_phys_src_dir = os.path.join(atm_root, "src", "physics", "ncar_ccpp")
+    source_search = [source_mods_dir, atm_phys_src_dir]
     # Find all metadata files, organize by scheme name
     all_scheme_files = _find_metadata_files(source_search, find_scheme_names)
 
@@ -449,8 +449,8 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
     for sdf in phys_suites_str.split(';'):
         sdf_path = _find_file(f"suite_{sdf}.xml", source_search)
         if not sdf_path:
-            emsg = "ERROR: Unable to find SDF for suite '{}'"
-            raise CamAutoGenError(emsg.format(sdf))
+            emsg = f"ERROR: Unable to find SDF for suite '{sdf}'"
+            raise CamAutoGenError(emsg)
         # End if
         sdfs.append(sdf_path)
         # Given an SDF, find all the schemes it calls
@@ -579,6 +579,30 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
         capgen_db = capgen(run_env, return_db=True)
     else:
         capgen_db = None
+    # end if
+
+    # Several SIMA dycores need some of the physics schemes
+    # located in the "utilities" CCPP physics directory,
+    # even if the actual physics suites don't need them.
+    # So go-ahead and copy all of the source code from
+    # there to the bld directory:
+    if do_gen_ccpp:
+        # Set CCPP physics "utilities" path
+        atm_phys_util_dir = os.path.join(atm_phys_src_dir, "utilities")
+
+        # Check that directory exists
+        if not os.path.isdir(atm_phys_util_dir):
+            #CAM-SIMA will likely not run without this, so raise an error
+            emsg = "ERROR: Unable to find CCPP physics utilities directory:\n"
+            emsg += f" {atm_phys_util_dir}\n Have you run 'checkout_externals'?"
+            raise CamAutoGenError(emsg)
+        # end if
+
+        # Copy all utility source files to the build directory
+        atm_phys_util_files = glob.glob(os.path.join(atm_phys_util_dir, "*.F90"))
+        for util_file in atm_phys_util_files:
+            shutil.copy2(util_file, physics_blddir)
+        # end for
     # end if
 
     if do_gen_ccpp or do_gen_nl:
--- a/cime_config/cam_autogen.py
+++ b/cime_config/cam_autogen.py
@@ -601,7 +601,7 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
         # Copy all utility source files to the build directory
         atm_phys_util_files = glob.glob(os.path.join(atm_phys_util_dir, "*.F90"))
         for util_file in atm_phys_util_files:
-            shutil.copy2(util_file, physics_blddir)
+            shutil.copy(util_file, physics_blddir)
         # end for
     # end if
 
--- a/src/control/cam_comp.F90
+++ b/src/control/cam_comp.F90
@@ -506,7 +506,7 @@ CONTAINS
       ! Dummy arguments
       type(runtime_options), intent(in) :: cam_runtime_opts
       ! Local variables
-      logical                                        :: is_const
+      logical                                        :: is_constituent
       integer                                        :: num_advect
       integer                                        :: const_idx
       integer                                        :: errflg
@@ -522,7 +522,7 @@ CONTAINS
       ! physics:
       call cam_ccpp_is_scheme_constituent(                                              &
            "water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water",                &
-           is_const, errflg, errmsg)
+           is_constituent, errflg, errmsg)
 
       if (errflg /= 0) then
          call endrun(subname//trim(errmsg), file=__FILE__, line=__LINE__)
@@ -531,7 +531,7 @@ CONTAINS
       !If not requested by the physics, then add water vapor to the
       !constituents object:
       !-------------------------------------------
-      if (.not. is_const) then
+      if (.not. is_constituent) then
 
          ! Allocate host_constituents object:
          allocate(host_constituents(1), stat=errflg)
--- a/src/control/cam_comp.F90
+++ b/src/control/cam_comp.F90
@@ -513,9 +513,12 @@ CONTAINS
       character(len=*), parameter :: subname = 'cam_register_constituents: '
 
       ! Register the constituents to find out what needs advecting
-      call host_constituents(1)%instantiate(std_name="specific_humidity",      &
-           long_name="Specific humidity", units="kg kg-1",                    &
-           vertical_dim="vertical_layer_dimension", advected=.true.,          &
+      call host_constituents(1)%instantiate( &
+           std_name="water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water",    &
+           long_name="water vapor mixing ratio w.r.t moist air and condensed_water", &
+           units="kg kg-1",                                                          &
+           vertical_dim="vertical_layer_dimension",                                  &
+           advected=.true.,                                                          &
            errcode=errflg, errmsg=errmsg)
       if (errflg /= 0) then
          call endrun(subname//trim(errmsg), file=__FILE__, line=__LINE__)
--- a/src/control/cam_comp.F90
+++ b/src/control/cam_comp.F90
@@ -50,11 +50,10 @@ module cam_comp
 
    logical  :: BFB_CAM_SCAM_IOP = .false.
 
-   ! Currently, the host (CAM) only adds water vapor (specific_humidity)
-   !    as a constituent.
-   ! Does this need to be a configurable variable?
-   integer, parameter :: num_host_advected = 1
-   type(ccpp_constituent_properties_t), target :: host_constituents(num_host_advected)
+   ! Currently, the host (CAM-SIMA) adds only water vapor (specific humidity)
+   ! as a constituent when not requested by the physics.  However, it is
+   ! unclear if this is actually necessary.
+   type(ccpp_constituent_properties_t), allocatable, target :: host_constituents(:)
 
    ! Private interface (here to avoid circular dependency)
    private :: cam_register_constituents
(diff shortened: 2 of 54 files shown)


nusbaume:
The test failures in python 3.12 are due to issues with ParamGen in CIME using a deprecated package.  I'll try to get that fixed and the CIME external updated, but if I don't get there in time I'll just remove python 3.12 from the testing options, at least for now.  

So feel free to ignore those failures when reviewing.  Thanks!

nusbaume:
All test failures have been fixed now, so this PR should be ready for review.  Thanks!

cacraigucar (changes requested):
Submitting current comments/requests.  I may find a few more, but wanted to send these while you might see them

mwaxmonsky (commented):
Just looking for some clarification on a few things but otherwise, looks great @nusbaume!!

src/control/cam_comp.F90
```diff
@@ -492,41 +491,86 @@ end subroutine cam_final
...
       use cam_ccpp_cap,              only: cam_ccpp_number_constituents
       use cam_ccpp_cap,              only: cam_model_const_properties
-      use cam_ccpp_cap,              only: cam_const_get_index
+      use cam_ccpp_cap,              only: cam_ccpp_is_scheme_constituent
 
       ! Dummy arguments
       type(runtime_options), intent(in) :: cam_runtime_opts
       ! Local variables
-      integer                                        :: index
+      logical                                        :: is_const
```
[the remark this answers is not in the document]
nusbaume: Replaced `is_const` variable with `is_constituent` to help make it more clear.

src/data/air_composition.F90
```diff
@@ -260,6 +265,9 @@ subroutine air_composition_init()
             thermodynamic_active_species_kc(icnst)  = kc1
             icnst = icnst + 1
             dry_species_num = dry_species_num + 1
+            !Notify constituent object that this species is
+            !thermodynamically active
+            call const_set_thermo_active(idx, .true.)
```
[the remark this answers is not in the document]
nusbaume: The default is `.false.`, which is specified in the constituents object itself:

https://github.com/peverwhee/ccpp-framework/blob/add_const_interface/src/ccpp_constituent_prop_mod.F90#L38

src/data/air_composition.F90
```diff
@@ -623,15 +677,14 @@ subroutine get_cp_1hd(tracer, inv_cp, cp, dp_dry, active_species_idx_dycore)
               (tracer(:,:,itrac) * factor(:,:))
       end do
 
-      if (dry_air_species_num == 0) then
-         sum_cp = thermodynamic_active_species_cp(0)
-      else
-         call get_cp_dry(tracer, idx_local, sum_cp, fact=factor)
-      end if
+      ! Get Cp for dry air:
```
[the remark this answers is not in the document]
nusbaume: I have updated the comment to say:

`! Get heat capacity at constant pressure (Cp) for dry air:`

src/data/ref_pres.F90
```diff
@@ -172,13 +171,11 @@ subroutine ref_pres_init(pref_edge_in, pref_mid_in, num_pr_lev_in)
       ! pref_edge_in
       call mark_as_initialized("reference_pressure_at_interface")
       ! pref_mid_in
-      call mark_as_initialized("reference_pressure")
+      call mark_as_initialized("reference_pressure_of_atmosphere_layer")
       ! pref_mid_norm
-      call mark_as_initialized("reference_pressure_normalized_by_surface_pressure")
+      call mark_as_initialized("reference_pressure_in_atmosphere_layer_normalized_by_reference_pressure")
```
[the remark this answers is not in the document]
nusbaume: You are definitely correct that there is a standard name bug here, but the error isn't with this standard name, but the one before it, which should be `reference_pressure_in_atmosphere_layer` instead of `reference_pressure_of_atmosphere_layer`.

This is because the `in_atmosphere_layer` version is already in the standard names dictionary and has been accepted by the scientists according to our spreadsheet.  I've thus gone ahead and replaced `reference_pressure_of_atmosphere_layer` with `reference_pressure_in_atmosphere_layer` in CAM-SIMA.

cime_config/cam_autogen.py
```diff
@@ -581,6 +581,30 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
...
+            #CAM-SIMA will likely not run without this, so raise an error
+            emsg = "ERROR: Unable to find CCPP physics utilities directory:\n"
+            emsg += f" {atm_phys_util_dir}\n Have you run 'checkout_externals'?"
+            raise CamAutoGenError(emsg)
+        # end if
+
+        # Copy all utility source files to the build directory
+        atm_phys_util_files = glob.glob(os.path.join(atm_phys_util_dir, "*.F90"))
+        for util_file in atm_phys_util_files:
+            shutil.copy2(util_file, physics_blddir)
```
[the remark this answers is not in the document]
nusbaume: You mean other than paranoia?  None that I can think of.  In fact I believe `shutil.copy` may be the better choice here, so I've gone ahead and changed the copy method to that.

src/dynamics/se/dp_coupling.F90
```diff
@@ -547,20 +550,24 @@ subroutine derived_phys_dry(cam_runtime_opts, phys_state, phys_tend)
...
+   use physics_types,     only: lagrangian_vertical
+   use physconst,         only: cpair, gravit, zvir, cappa
+   use cam_thermo,        only: cam_thermo_update
+   use physics_types,     only: cpairv, rairv, zvirv
+   use physics_grid,      only: columns_on_task
+   use shr_const_mod,     only: shr_const_rwv
+   use geopotential_temp, only: geopotential_temp_run
+   use static_energy,     only: update_dry_static_energy_run
+!   use qneg,              only: qneg_run
 !   use check_energy,   only: check_energy_timestep_init
```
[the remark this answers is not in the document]
nusbaume: Normally yes, but currently the SE dycore in CAM-SIMA is still missing a lot of required features, so having this commented-out code is a good reminder that we'll need to re-add this particular feature at some point in the future before the model is ready for scientific use.

However, if you don't think that is a good reason then I am happy to remove it, as the compiler will always notify me that it is missing as well.

src/dynamics/se/dp_coupling.F90
```diff
@@ -726,29 +747,27 @@ subroutine derived_phys_dry(cam_runtime_opts, phys_state, phys_tend)
...
+   ! Please note this cannot be used until the 'qmin'
+   ! array is publicly provided by the CCPP constituent object. -JN
 #if 0
-      ! Ensure tracers are all positive
-      call qneg3('D_P_COUPLING',lchnk  ,ncol    ,pcols   ,pver    , &
-           1, num_advected, qmin  ,const_data_ptr)
+   call qneg_run('D_P_COUPLING', columns_on_task, pver, &
+                 qmin, const_data_ptr,       &
+                 errflg, errmsg)
 #endif
```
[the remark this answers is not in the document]
nusbaume: It's because this code doesn't work yet in CAM-SIMA (as it needs a new CCPP constituents host-side "get" method), but we will need it to work in order for SIMA to match CAM's scientific results when running with the SE dycore.  The `#if 0` code is designed to both turn this code off for now, and be really noticeable so that we remember to fix it in the future.  If you think there might be a better way to do accomplish that goal then just let me know!

src/control/cam_comp.F90
```diff
@@ -536,9 +580,25 @@ subroutine cam_register_constituents(cam_runtime_opts)
...
 
-      ! Finally, initialize the constituents module
+      ! Initialize the constituents module
       call cam_constituents_init(const_props, num_advect)
 
+      ! Finally, for CAM moist physics, one will need to use a
+      ! non-zero minimum value for water vapor, so update that
+      ! value here:
+      !-------------------------------------------
+      if (phys_suite_name /= 'held_suarez_1994') then !Held-Suarez is "dry" physics
```
peverwhee, line 590: would it make sense to have some sort of is_dry logical for schemes? Is it possible that we'll have other dry physics? Or is this a PFFU (problem for future us)

src/dynamics/se/dp_coupling.F90
```diff
@@ -547,20 +550,24 @@ subroutine derived_phys_dry(cam_runtime_opts, phys_state, phys_tend)
...
+   use ccpp_constituent_prop_mod, only: ccpp_constituent_prop_ptr_t
+   use cam_ccpp_cap,      only: cam_constituents_array
+   use cam_ccpp_cap,      only: cam_model_const_properties
+   use cam_constituents,  only: const_get_index
+   use physics_types,     only: lagrangian_vertical
+   use physconst,         only: cpair, gravit, zvir, cappa
+   use cam_thermo,        only: cam_thermo_update
+   use physics_types,     only: cpairv, rairv, zvirv
+   use physics_grid,      only: columns_on_task
+   use shr_const_mod,     only: shr_const_rwv
```
peverwhee, line 562: I don't think this is used in this subroutine

src/dynamics/utils/hycoef.F90
```diff
@@ -100,8 +99,7 @@ subroutine hycoef_init(file, psdry)
    !-----------------------------------------------------------------------
 
    ! Initalize reference pressures:
```
peverwhee, line 101: teeny nitpick -
```suggestion
   ! Initalize reference pressure:
```

src/physics/utils/physics_data.F90
```diff
@@ -348,19 +349,24 @@ subroutine check_field_2d(file, var_names, timestep, current_value,        &
       real(kind_phys)                  :: diff
       integer                          :: col
       integer                          :: ierr      !For MPI
+      integer                          :: mpi_stat(mpi_status_size) !For MPI
       real(kind_phys), allocatable     :: buffer(:)
       integer                          :: diff_count
-      real(kind_phys)                  :: max_diff
-      real(kind_phys)                  :: max_diff_gl
+      real(kind_phys)                  :: max_diff(2,1)    !Stores the local max diff and its MPI rank
```
peverwhee, line 355: why the second dimension (for this and max_diff_gl)? Some MPI necessity?

src/physics/utils/physics_data.F90
```diff
@@ -501,36 +577,54 @@ subroutine check_field_3d(file, var_names, vcoord_name, timestep,          &
...
       if (slen > indent_level) then
          write(iulog, '(a)') trim(stdname)
          slen = 0
       end if
-      write(fmt_str, '(a,i0,a)') "(1x,a,t",indent_level+1,",1x,i7,2x,e8.2)"
-      write(iulog, fmt_str) stdname(1:slen), diff_count, max_diff
+
+      !Write out difference and index valuesa:
+      if (present(max_diff_lev)) then
+         write(index_str, '(a,i0,a,i0,a,i0,a)') "(",max_diff_rank,",",max_diff_col,",",max_diff_lev,"}"
```
peverwhee, line 622: why the different types of parenthesis/brackets?

peverwhee (changes requested):
looks good; just a couple questions/comments.

src/control/cam_comp.F90
```diff
@@ -492,41 +491,86 @@ end subroutine cam_final
...
+      !constituents object:
+      !-------------------------------------------
+      if (.not. is_constituent) then
+
+         ! Allocate host_constituents object:
+         allocate(host_constituents(1), stat=errflg)
+         call check_allocate(errflg, subname, 'host_constituents(1)',                   &
+                             file=__FILE__, line=__LINE__)
+
+         ! Register the constituents to find out what needs advecting
```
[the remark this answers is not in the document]
nusbaume: Comment improved as suggested.

src/control/cam_comp.F90
nusbaume: I don't personally think we should worry about adding a logical here, for two reasons:

1.  Only Held-Suarez and "adiabatic" (which basically means no physics at all) have the minimum set to zero.  All other CAM physics schemes use the default minimum value.  So I feel like adding a logical and a framework to support that logical would probably not be worth the effort (at least for CAM physics suites).

2.  Ideally I don't want CAM-SIMA to add water vapor (or any constituent) by default.  Instead it should be up to the physics, chemistry, and dycore to add whatever constituents they need.  This means that eventually I hope this whole code block will go away.  Given this, I want to put the bare minimum needed into this particular set of routines.  As an FYI, there is already an issue open (issue #208) for this future development plan.

Of course if you disagree with either of these points, or think there would be additional benefits to having an `is_dry` logical, then just let me know!

src/dynamics/se/dp_coupling.F90
nusbaume: Good catch!  Non-used statement removed.

src/dynamics/utils/hycoef.F90
nusbaume: Louse egg removed.

src/physics/utils/physics_data.F90
nusbaume: Good question!  It's because I don't know how to read MPI documentation properly.  Extraneous dimension removed.

src/physics/utils/physics_data.F90
nusbaume: There are two possibilities.  One is that I have a deep, secret, subconscious hatred of parenthetical symmetry.  The other is that I made a typo and then copied that typo.  I'll leave it up to you to decide which is the most likely option.

Also, parenthesis are now both the same.

peverwhee (approved):
thanks Jesse!

I also resolved @cacraigucar 's comments that you addressed (but left the comments that Cheryl may want to see the answer to when she returns).

src/control/cam_comp.F90
gold2718: Is there a replacement for the `moist_physics` logical that is used, for example, in cam_diagnostics.F90?

nusbaume:
@mwaxmonsky if you get the chance can you approve or re-review this PR?  I am trying to avoid breaking too many of my own rules here.

src/control/cam_comp.F90
nusbaume: @gold2718 Not yet, but since we don't have the new history infrastructure plugged in yet I figure we can tackle that particular concern later (but I definitely agree we need some general way of allowing physics suites to tell CAM-SIMA what standard outputs should or should not be included by default).

nusbaume:
@mwaxmonsky it looks like you resolved your concerns earlier, and since I am hoping to have this repo ready for the presentation in the next 40 minutes I am going to go ahead and break my own rules and get this merged in.  Of course feel free to still review the code if you want and let me know if there is anything problematic (as we can always open another PR).  Thanks!

PR #244, opened 2023-12-30
Stub interface for MPAS dynamical core

kuanchihwang:
This PR introduces stub interface for MPAS dynamical core. This stub interface currently does nothing (i.e., can compile; cannot run), but it provides a fresh start from scratch for actual functionalities to be incrementally implemented later.

PR #245, opened 2024-01-08
Integrate MPAS build infrastructure with CIME

kuanchihwang:
This PR brings integration of MPAS build infrastructure with CIME. If MPAS dynamical core is selected (determined from model grid specifications), it will be built as a part of the usual CIME build process (i.e., `case.build`).

The approach taken here is to reuse upstream MPAS build infrastructure as much as possible. This way, it would be easier to keep in sync with future upstream changes. However, by default, MPAS is built as a standalone executable. A supplemental `Makefile` is therefore needed to enable MPAS to be built as a static library, which is then linked into the host model executable.

This PR should be considered *after* #244.

### DIFF
--- a/cime_config/buildlib
+++ b/cime_config/buildlib
@@ -141,6 +141,11 @@ def _build_cam():
         #-------------------------------------------------------
         # build the library
         #-------------------------------------------------------
+
+        # If dynamical core is MPAS, setup its build infrastructure so it can be built along with CAM below.
+        if dycore == "mpas":
+            _setup_mpas(case)
+
         complib = os.path.join(libroot, "libatm.a")
         makefile = os.path.join(casetools, "Makefile")
 
@@ -156,22 +161,13 @@ def _build_cam():
         _LOGGER.info("%s: \n\n output:\n %s \n\n err:\n\n%s\n", cmd, out, err)
         expect(retcode == 0, f"Command {cmd} failed with rc={retcode}")
 
-        # If dynamical core is MPAS, build it after CAM due to dependencies.
-        if dycore == "mpas":
-            _build_mpas(case)
-
-def _build_mpas(case: Case) -> None:
+def _setup_mpas(case: Case) -> None:
     """
-    Build MPAS by invoking its build infrastructure.
+    Setup MPAS build infrastructure.
     """
 
-    case_root    = os.path.normpath(case.get_value("CASEROOT"))
     atm_src_root = os.path.normpath(case.get_value("COMP_ROOT_DIR_ATM"))
     atm_bld_root = os.path.normpath(os.path.join(case.get_value("EXEROOT"), "atm", "obj"))
-    lib_root     = os.path.normpath(os.path.join(case.get_value("EXEROOT"), "lib"))
-
-    gmake   = case.get_value("GMAKE")
-    gmake_j = case.get_value("GMAKE_J")
 
     mpas_dycore_src_root = os.path.join(atm_src_root, "src", "dynamics", "mpas", "dycore", "src")
     mpas_dycore_bld_root = os.path.join(atm_bld_root, "mpas")
@@ -188,23 +184,9 @@ def _build_mpas(case: Case) -> None:
         raise FileExistsError(1, "Unexpected file", mpas_dycore_bld_root)
 
     shutil.copytree(mpas_dycore_src_root, mpas_dycore_bld_root, copy_function=_copy2_as_needed, dirs_exist_ok=True)
-    _copy2_as_needed(os.path.normpath(os.path.join(mpas_dycore_src_root, "../../Makefile.in.CESM")), os.path.join(mpas_dycore_bld_root, "Makefile.in.CESM"))
-
-    # Invoke makefile target `libmpas-prepare`.
-    mpas_dycore_make_command  = f'{gmake} -C {mpas_dycore_bld_root} libmpas-prepare ESM="CESM" CASEROOT="{case_root}" LIBROOT="{lib_root}" '
-    mpas_dycore_make_command += f'FCINCLUDES="-I{atm_bld_root}"'
-
-    (sta, out, err) = run_cmd(mpas_dycore_make_command, from_dir=mpas_dycore_bld_root)
-    _LOGGER.info("* Command %s:\n\n  Output:\n%s\n\n  Error:\n%s\n\n", mpas_dycore_make_command, out, err)
-    expect(sta == 0, f"Command {mpas_dycore_make_command} failed with status {sta}")
-
-    # Invoke makefile target `libmpas-build`.
-    mpas_dycore_make_command  = f'{gmake} -C {mpas_dycore_bld_root} -j {gmake_j} libmpas-build ESM="CESM" CASEROOT="{case_root}" LIBROOT="{lib_root}" '
-    mpas_dycore_make_command += f'FCINCLUDES="-I{atm_bld_root}"'
-
-    (sta, out, err) = run_cmd(mpas_dycore_make_command, from_dir=mpas_dycore_bld_root)
-    _LOGGER.info("* Command %s:\n\n  Output:\n%s\n\n  Error:\n%s\n\n", mpas_dycore_make_command, out, err)
-    expect(sta == 0, f"Command {mpas_dycore_make_command} failed with status {sta}")
+    shutil.move(os.path.join(mpas_dycore_bld_root, "Makefile"), os.path.join(mpas_dycore_bld_root, "Makefile.CESM"))
+    _copy2_as_needed(os.path.normpath(os.path.join(mpas_dycore_src_root, os.pardir, os.pardir, "Makefile")), os.path.join(mpas_dycore_bld_root, "Makefile"))
+    _copy2_as_needed(os.path.normpath(os.path.join(mpas_dycore_src_root, os.pardir, os.pardir, "Makefile.in.CESM")), os.path.join(mpas_dycore_bld_root, "Makefile.in.CESM"))
 
 def _copy2_as_needed(src: str, dst: str) -> None:
     """
--- a/cime_config/buildlib
+++ b/cime_config/buildlib
@@ -187,9 +187,8 @@ def _build_mpas(case: Case) -> None:
     if os.path.isfile(mpas_dycore_bld_root):
         raise FileExistsError(1, "Unexpected file", mpas_dycore_bld_root)
 
-    if not os.path.isdir(mpas_dycore_bld_root):
-        shutil.copytree(mpas_dycore_src_root, mpas_dycore_bld_root)
-        shutil.copy2(os.path.normpath(os.path.join(mpas_dycore_src_root, "../../Makefile.in.CESM")), os.path.join(mpas_dycore_bld_root, "Makefile.in.CESM"))
+    shutil.copytree(mpas_dycore_src_root, mpas_dycore_bld_root, copy_function=_copy2_as_needed, dirs_exist_ok=True)
+    _copy2_as_needed(os.path.normpath(os.path.join(mpas_dycore_src_root, "../../Makefile.in.CESM")), os.path.join(mpas_dycore_bld_root, "Makefile.in.CESM"))
 
     # Invoke makefile target `libmpas-prepare`.
     mpas_dycore_make_command  = f'{gmake} -C {mpas_dycore_bld_root} libmpas-prepare ESM="CESM" CASEROOT="{case_root}" LIBROOT="{lib_root}" '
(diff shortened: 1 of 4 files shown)


nusbaume:
@kuanchihwang just letting you know that I'll make sure to review this PR once PR #244 has been merged (as you suggested above).  Thanks!

cime_config/buildlib
```diff
@@ -156,6 +156,57 @@ def _build_cam():
...
+
+    if not os.path.isdir(mpas_dycore_src_root):
+        raise FileNotFoundError(1, "No such directory", mpas_dycore_src_root)
+
+    # MPAS supports "in-source" build only. Copy source code to `mpas_dycore_bld_root` to avoid polluting `mpas_dycore_src_root`.
+    if os.path.isfile(mpas_dycore_bld_root):
+        raise FileExistsError(1, "Unexpected file", mpas_dycore_bld_root)
+
+    if not os.path.isdir(mpas_dycore_bld_root):
+        shutil.copytree(mpas_dycore_src_root, mpas_dycore_bld_root)
```
mgduda, line 191: How will this work if we modify an MPAS-A dycore source file in the CAM-SIMA source tree? If we subsequently run `case.build` in our case directory, will the modifications to the source be correctly incorporated?

cime_config/config_component.xml
```diff
@@ -303,11 +304,12 @@
 
   <entry id="CAM_LINKED_LIBS">
     <type>char</type>
-    <valid_values></valid_values>
+    <valid_values>-lmpas</valid_values>
```
nusbaume, line 307: I might vote to not worry about updating `valid_values` here, as I could imagine all sorts of different libraries that might be linked in the future, and we probably don't want to require future users/developers to have to update this line.  

Plus if someone mistakenly tries to link the wrong library it will probably either be ignored or cause an obvious compiler or linker error at build time.

cime_config/config_component.xml
```diff
@@ -303,11 +304,12 @@
...
     <default_value>-lmusica -ljsonfortran</default_value>
     <values match="last">
-      <!--Turn off for physics testbed -->
-      <value compset="_CAM%PHYSTEST"> </value>
+      <!-- Turn off for physics testbed -->
+      <value compset="_CAM%PHYSTEST"></value>
+      <value grid="a%mpasa[0-9]+">-lmpas</value>
```
nusbaume, line 312: This is fine as long as there are no immediate plans to run MPAS in CAM-SIMA with chemistry.  Otherwise we'll need to modify the `<value match="last">` line above to be `<value match="last" modifier="additive">`.

cime_config/config_component.xml
```diff
@@ -103,10 +103,11 @@
 
   <entry id="CAM_DYCORE">
     <type>char</type>
-    <valid_values>eul,fv,se,none</valid_values>
+    <valid_values>eul,fv,mpasa,se,none</valid_values>
```
nusbaume, line 106: If you use the CIME Makefile method then I believe the `CAM_DYCORE` variable should be set to `mpas` instead of  `mpasa`.

src/dynamics/mpas/Makefile.in.CESM
```diff
@@ -0,0 +1,110 @@
...
+                         -DMPAS_BUILD_TARGET="$(BUILD_TARGET)" \
+                         -DMPAS_CAM_DYCORE \
+                         -DMPAS_EXE_NAME="$(EXE_NAME)" \
+                         -DMPAS_EXTERNAL_ESMF_LIB \
+                         -DMPAS_GIT_VERSION="$(GIT_VERSION)" \
+                         -DMPAS_NAMELIST_SUFFIX="$(NAMELIST_SUFFIX)" \
+                         -DMPAS_NATIVE_TIMERS \
+                         -DMPAS_NO_ESMF_INIT \
+                         -DMPAS_PIO_SUPPORT \
+                         -DUSE_PIO2
```
nusbaume, line 48: If you use CIME's `Makefile` then I don't believe you need this `-DUSE_PIO2` line here.

src/dynamics/mpas/Makefile.in.CESM
```diff
@@ -0,0 +1,110 @@
...
+export CPPINCLUDES    := $(strip $(CPPINCLUDES))
+export CXX            := $(strip $(MPICXX))
+export CXXFLAGS       := $(strip $(CXXFLAGS))
+export FC             := $(strip $(MPIFC))
+export FCINCLUDES     := $(strip $(FCINCLUDES))
+export FFLAGS         := $(strip $(FC_AUTO_R8) $(FFLAGS) $(FREEFLAGS))
```
nusbaume, line 54: If you use the CIME Makefile then I believe `FFLAGS` and `FREEFLAGS` are already combined within the `FFLAGS` variable.

nusbaume (changes requested):
Thanks for the PR @kuanchihwang !  However I believe many of these changes may be unnecssary as the top-level CESM Makefile in CIME already has the setup needed to build the MPAS-A dycore as a library.  For example see here:

https://github.com/ESMCI/cime/blob/master/CIME/Tools/Makefile#L519

and here:

https://github.com/ESMCI/cime/blob/master/CIME/Tools/Makefile#L556

So I think in CAM-SIMA you only have to create the relevant Makefile and associated paths/linking commands (which you are doing in the PR already).  Everything else can theoretically be removed, which explains my comments below.

Of course if you and @mgduda would prefer to move away from the CIME Makefile method then please let me know and I'll change my review.  Thanks!

cime_config/buildlib
kuanchihwang: This should be addressed by https://github.com/ESCOMP/CAM-SIMA/pull/245/commits/0959d93b2baf5bd8ed98d9b77fd569c7df432059 below.

cime_config/buildlib
```diff
@@ -156,6 +156,57 @@ def _build_cam():
         _LOGGER.info("%s: \n\n output:\n %s \n\n err:\n\n%s\n", cmd, out, err)
         expect(retcode == 0, f"Command {cmd} failed with rc={retcode}")
 
+        # If dynamical core is MPAS, build it after CAM due to dependencies.
+        if dycore == "mpas":
+            _build_mpas(case)
+
+def _build_mpas(case: Case) -> None:
+    """
+    Build MPAS by invoking its build infrastructure.
```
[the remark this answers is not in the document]
kuanchihwang: This should be addressed by https://github.com/ESCOMP/CAM-SIMA/pull/245/commits/4f47388f5bd1badc05446f9e4163006fe9a0e96c below. Function `_setup_mpas` in `cime_config/buildlib` now only arranges source code and makefiles in a way that MPAS can be built along with CAM by CIME.

cime_config/buildlib
```diff
@@ -156,6 +156,57 @@ def _build_cam():
...
+    """
+
+    case_root    = os.path.normpath(case.get_value("CASEROOT"))
+    atm_src_root = os.path.normpath(case.get_value("COMP_ROOT_DIR_ATM"))
+    atm_bld_root = os.path.normpath(os.path.join(case.get_value("EXEROOT"), "atm", "obj"))
+    lib_root     = os.path.normpath(os.path.join(case.get_value("EXEROOT"), "lib"))
+
+    gmake   = case.get_value("GMAKE")
+    gmake_j = case.get_value("GMAKE_J")
```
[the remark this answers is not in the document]
kuanchihwang: This should be addressed by https://github.com/ESCOMP/CAM-SIMA/commit/4f47388f5bd1badc05446f9e4163006fe9a0e96c below. The relevant code has been deleted.

cime_config/buildlib
```diff
@@ -156,6 +156,57 @@ def _build_cam():
...
+    if not os.path.isdir(mpas_dycore_src_root):
+        raise FileNotFoundError(1, "No such directory", mpas_dycore_src_root)
+
+    # MPAS supports "in-source" build only. Copy source code to `mpas_dycore_bld_root` to avoid polluting `mpas_dycore_src_root`.
+    if os.path.isfile(mpas_dycore_bld_root):
+        raise FileExistsError(1, "Unexpected file", mpas_dycore_bld_root)
+
+    if not os.path.isdir(mpas_dycore_bld_root):
+        shutil.copytree(mpas_dycore_src_root, mpas_dycore_bld_root)
+        shutil.copy2(os.path.normpath(os.path.join(mpas_dycore_src_root, "../../Makefile.in.CESM")), os.path.join(mpas_dycore_bld_root, "Makefile.in.CESM"))
```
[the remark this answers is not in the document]
kuanchihwang: Thanks! I did not know about `os.pardir`. Incorporated the changes in https://github.com/ESCOMP/CAM-SIMA/commit/4f47388f5bd1badc05446f9e4163006fe9a0e96c.

cime_config/buildlib
```diff
@@ -156,6 +156,57 @@ def _build_cam():
...
+    _LOGGER.info("* Command %s:\n\n  Output:\n%s\n\n  Error:\n%s\n\n", mpas_dycore_make_command, out, err)
+    expect(sta == 0, f"Command {mpas_dycore_make_command} failed with status {sta}")
+
+    # Invoke makefile target `libmpas-build`.
+    mpas_dycore_make_command  = f'{gmake} -C {mpas_dycore_bld_root} -j {gmake_j} libmpas-build ESM="CESM" CASEROOT="{case_root}" LIBROOT="{lib_root}" '
+    mpas_dycore_make_command += f'FCINCLUDES="-I{atm_bld_root}"'
+
+    (sta, out, err) = run_cmd(mpas_dycore_make_command, from_dir=mpas_dycore_bld_root)
+    _LOGGER.info("* Command %s:\n\n  Output:\n%s\n\n  Error:\n%s\n\n", mpas_dycore_make_command, out, err)
+    expect(sta == 0, f"Command {mpas_dycore_make_command} failed with status {sta}")
```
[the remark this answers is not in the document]
kuanchihwang: This should be addressed by https://github.com/ESCOMP/CAM-SIMA/commit/4f47388f5bd1badc05446f9e4163006fe9a0e96c below. Function `_setup_mpas` in `cime_config/buildlib` now only arranges source code and makefiles in a way that MPAS can be built along with CAM by CIME.

cime_config/config_component.xml
kuanchihwang: Of course. I reverted this change in https://github.com/ESCOMP/CAM-SIMA/pull/245/commits/3a9b8ad2c0ceecf740a60c40dfade1c78eff1f74.

cime_config/config_component.xml
kuanchihwang: Of course. I made the suggested change in https://github.com/ESCOMP/CAM-SIMA/commit/3a9b8ad2c0ceecf740a60c40dfade1c78eff1f74.

cime_config/config_component.xml
kuanchihwang: Since I switched back to use CIME Makefile for building, this change has been reverted in https://github.com/ESCOMP/CAM-SIMA/pull/245/commits/38d0de2f3891a56c4d1a8d00d2f9234fc91f692a.

src/dynamics/mpas/Makefile.in.CESM
kuanchihwang: Should be addressed in https://github.com/ESCOMP/CAM-SIMA/pull/245/commits/9842660369bcbead473656860787ec430a28194a. `PIODEF` is defined by CIME Makefile and its value can be empty or `-DUSE_PIO2`. Use it here instead.

src/dynamics/mpas/Makefile.in.CESM
kuanchihwang: Should be addressed in https://github.com/ESCOMP/CAM-SIMA/commit/9842660369bcbead473656860787ec430a28194a. Rely on CIME Makefile to handle this.

cime_config/buildlib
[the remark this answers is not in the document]
kuanchihwang: The top-level CIME Makefile (i.e., `cime/CIME/Tools/Makefile`) does not handle this. It queries the value of `CAM_DYCORE` in `$CASEROOT/env_build.xml` [[1]](https://github.com/ESMCI/cime/blob/d2ec81a4e5ee4bc558c46f31b407c02afd759b6b/CIME/Tools/Makefile#L513), and if the value is `mpas`, it just invokes the Makefile in `$EXEROOT/atm/obj/mpas` [[2]](https://github.com/ESMCI/cime/blob/d2ec81a4e5ee4bc558c46f31b407c02afd759b6b/CIME/Tools/Makefile#L575).

In CAM, the Makefile in `$EXEROOT/atm/obj/mpas` is copied from `src/dynamics/mpas` [[3]](https://github.com/ESCOMP/CAM/blob/452cb5b7f5efa7f421ff6888b0f49c955e0b9579/bld/configure#L2043). If you inspect it [[4]](https://github.com/ESCOMP/CAM/blob/452cb5b7f5efa7f421ff6888b0f49c955e0b9579/src/dynamics/mpas/Makefile), you can find that it uses `VPATH` to point to MPAS source code. However, the build targets, prerequisites, and recipes are listed manually. This seems like a maintenance burden if the upstream changes how MPAS is built.

In CAM-SIMA here, I attempt to delegate the work of building MPAS to its upstream build infrastructure, which supports "in-source" build only. This is the reason why MPAS source code needs to be copied to `$EXEROOT/atm/obj/mpas`. This way, we no longer have to maintain the relationship among build targets, prerequisites, and recipes ourselves.

The chain of events:

1. In `cime_config/buildlib`, function `_build_cam` calls function `_setup_mpas`.
2. Function `_setup_mpas` arranges MPAS source code and makefiles [[5]](https://github.com/ESCOMP/CAM-SIMA/blob/3a9b8ad2c0ceecf740a60c40dfade1c78eff1f74/cime_config/buildlib#L186-L189) according to the following table.

| Source                                           | Destination                              |
|--------------------------------------------------|------------------------------------------|
| `src/dynamics/mpas/dycore/src/*`                 | `$EXEROOT/atm/obj/mpas/*`                |
| `src/dynamics/mpas/dycore/src/Makefile`          | `$EXEROOT/atm/obj/mpas/Makefile.CESM`    |
| `src/dynamics/mpas/Makefile`                     | `$EXEROOT/atm/obj/mpas/Makefile`         |
| `src/dynamics/mpas/Makefile.in.CESM`             | `$EXEROOT/atm/obj/mpas/Makefile.in.CESM` |

3. Function `_build_cam` invokes the top-level CIME Makefile.
4. The top-level CIME Makefile invokes `$EXEROOT/atm/obj/mpas/Makefile`.
5. `$EXEROOT/atm/obj/mpas/Makefile` invokes `$EXEROOT/atm/obj/mpas/Makefile.CESM`. In the process, `$EXEROOT/atm/obj/mpas/Makefile.in.CESM` gets `include`d [[6]](https://github.com/MPAS-Dev/MPAS-Model/blob/cd652c32898f2b13e54df275c2944d4d04dfa451/src/Makefile#L5).
6. Upstream build infrastructure handles the work of building MPAS for us.

Actually, we can consider upstreaming `src/dynamics/mpas/Makefile.in.CESM` to MPAS repository. This is similar to what E3SM has done before [[7]](https://github.com/MPAS-Dev/MPAS-Model/blob/cd652c32898f2b13e54df275c2944d4d04dfa451/src/Makefile.in.E3SM).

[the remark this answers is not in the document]
nusbaume: @kuanchihwang thanks for double-checking for me, and for the explanation of your new workflow!  This all makes sense to me, and I agree your method is likely much more maintainable then a hand-modified Makefile on the CAM-SIMA side.

I'll leave the question of where the `Makefile.in.CESM` should live up to you and @mgduda.  I don't think I really have a preference on my end.  Thanks!

nusbaume (approved):
Everything looks good to me now.  Thanks @kuanchihwang!

mgduda (approved):
Looks good to me!

PR #249, opened 2024-02-15
Comment out linked libs for FPHYStest compset

peverwhee:
Temporary fix; this means we'll have to do ./xmlchange CAM_LINKED_LIBS="" before trying to build an FPHYStest case.


nusbaume (approved):

nusbaume:
@peverwhee @cacraigucar Courtney's branch should now allow you to run `create_newcase` and not have to do the extra `xmlchange CAM_LINKED_LIBS` command when running with the `FPHYStest` compset.  Of course if it doesn't work for you please let me know.  Thanks!

cacraigucar (approved):
This change allows me to successfully run a kessler test

PR #267, opened 2024-06-06
Implement and finish MPAS dynamical core initialization

kuanchihwang:
This PR implements the `dyn_init` subroutine, which is responsible for initializing MPAS dynamical core by one of the following methods:

1. Setting analytic initial condition.
2. Reading initial condition from a file.
3. Restarting from a file.

This should be the last step of MPAS dynamical core initialization. After this PR, MPAS dynamical core is ready for time integration.

### Implementation notes:

* More elaborate error checking for the allocation of constituents.
* Support for deferring the definition of constituents until run-time. There are no longer assumptions about the index order or position of constituents.
* Accessor functions/subroutines for querying:
  * Local mesh dimensions.
  * Constituent names and indexes.
  * Mapping between constituent indexes and MPAS scalar indexes.
* Enhancements that enable constructing an arbitrary list of variables to input/output data to/from MPAS dynamical core.
* Enhancements and fixes for handling ordinary variables and variable arrays during input.
* Finish MPAS dynamical core initialization.

### Test steps:

```bash
"<path-to>/CAM-SIMA/cime/scripts/create_newcase" --case "<case-name>" --compset FKESSLER --project "<project-id>" --res mpasa480_mpasa480 --run-unsupported
cd "<case-name>"
./case.setup
./xmlchange CAM_LINKED_LIBS="-lmpas"
./xmlchange DEBUG="TRUE"
./xmlchange DOUT_S="FALSE"
# Add the following to `user_nl_cam`:
# * `config_block_decomp_file_prefix = '${DIN_LOC_ROOT}/atm/cam/inic/mpas/mpasa480.graph.info.part.'`.
# * `config_dt = 1800.0`.
./case.build
./case.submit
```

Observe log entries similar to the following in `atm.log.<job-id>.<date>-<time>`:

```
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_define_scalar entered
dyn_mpas_debug_print (0): Constituent index 1
dyn_mpas_debug_print (0):     Constituent name: cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water
dyn_mpas_debug_print (0):     Is water species: T
dyn_mpas_debug_print (0):     Index mapping from constituent to MPAS scalar: 1 -> 1
dyn_mpas_debug_print (0): Constituent index 2
dyn_mpas_debug_print (0):     Constituent name: rain_mixing_ratio_wrt_moist_air_and_condensed_water
dyn_mpas_debug_print (0):     Is water species: T
dyn_mpas_debug_print (0):     Index mapping from constituent to MPAS scalar: 2 -> 2
dyn_mpas_debug_print (0): Constituent index 3
dyn_mpas_debug_print (0):     Constituent name: water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water
dyn_mpas_debug_print (0):     Is water species: T
dyn_mpas_debug_print (0):     Index mapping from constituent to MPAS scalar: 3 -> 3
dyn_mpas_debug_print (0): MPAS scalar index 1
dyn_mpas_debug_print (0):     MPAS scalar name: cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water
dyn_mpas_debug_print (0):     Is water species: T
dyn_mpas_debug_print (0):     Index mapping from MPAS scalar to constituent: 1 -> 1
dyn_mpas_debug_print (0): MPAS scalar index 2
dyn_mpas_debug_print (0):     MPAS scalar name: rain_mixing_ratio_wrt_moist_air_and_condensed_water
dyn_mpas_debug_print (0):     Is water species: T
dyn_mpas_debug_print (0):     Index mapping from MPAS scalar to constituent: 2 -> 2
dyn_mpas_debug_print (0): MPAS scalar index 3
dyn_mpas_debug_print (0):     MPAS scalar name: water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water
dyn_mpas_debug_print (0):     Is water species: T
dyn_mpas_debug_print (0):     Index mapping from MPAS scalar to constituent: 3 -> 3
dyn_mpas_debug_print (0): index_qv = 3
dyn_mpas_debug_print (0): moist_start = 1
dyn_mpas_debug_print (0): moist_end = 3
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_define_scalar completed
dyn_debug_print (0): Calling check_topography_data
dyn_debug_print (0): Topography file is not used
dyn_debug_print (0): Calling set_analytic_initial_condition
dyn_debug_print (0): Preparing to set analytic initial condition
dyn_debug_print (0): Setting MPAS state "u"
           U initialized by "BC_WAV_SET_IC"
           V initialized by "BC_WAV_SET_IC"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_compute_edge_wind entered
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "uReconstructZonal"
dyn_mpas_debug_print (0): Exchanging halo layers for "uReconstructZonal"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "uReconstructMeridional"
dyn_mpas_debug_print (0): Exchanging halo layers for "uReconstructMeridional"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "u"
dyn_mpas_debug_print (0): Exchanging halo layers for "u"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_compute_edge_wind completed
dyn_debug_print (0): Setting MPAS state "w"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "w"
dyn_mpas_debug_print (0): Exchanging halo layers for "w"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_debug_print (0): Setting MPAS state "scalars"
dyn_debug_print (0): Conversion is needed and applied for water vapor mixing ratio
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "scalars"
dyn_mpas_debug_print (0): Exchanging halo layers for "scalars"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_debug_print (0): Setting MPAS state "rho" and "theta"
 BC_WAV_SET_IC: Model top (in km) is at z=    41.426919000000005     
           PS initialized by "BC_WAV_SET_IC"
           T initialized by "BC_WAV_SET_IC"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "rho"
dyn_mpas_debug_print (0): Exchanging halo layers for "rho"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "theta"
dyn_mpas_debug_print (0): Exchanging halo layers for "theta"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_debug_print (0): Setting MPAS state "rho_base" and "theta_base"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "rho_base"
dyn_mpas_debug_print (0): Exchanging halo layers for "rho_base"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo entered
dyn_mpas_debug_print (0): Inquiring field information for "theta_base"
dyn_mpas_debug_print (0): Exchanging halo layers for "theta_base"
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_exchange_halo completed
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_init_phase4 entered
dyn_mpas_debug_print (0): Coupling time interval is   1800.000000 seconds
dyn_mpas_debug_print (0): Time step is   1800.000000 seconds
dyn_mpas_debug_print (0): Setting up OpenMP threading
dyn_mpas_debug_print (0): Setting up dimensions
dyn_mpas_debug_print (0): Initializing time levels
dyn_mpas_debug_print (0): Initializing atmospheric variables
dyn_mpas_debug_print (0): Initializing dynamics
dyn_mpas_debug_print (0): dyn_mpas_subdriver::dyn_mpas_init_phase4 completed
dyn_mpas_debug_print (0): Successful initialization of MPAS dynamical core
```


nusbaume (changes requested):
Exciting to see that we can now fully initialize the MPAS dycore in CAM-SIMA, good work!  

I do have some change requests though.  Of course please let me know if you have any questions or concerns with any of my comments or suggestions. Thanks!

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -1,28 +1,47 @@
...
+    use cam_field_read, only: cam_read_field
+    use cam_grid_support, only: cam_grid_get_latvals, cam_grid_get_lonvals, cam_grid_id
+    use cam_initfiles, only: initial_file_get_id, topo_file_get_id
     use cam_instance, only: atm_id
     use cam_logfile, only: debug_output, debugout_none, iulog
+    use cam_pio_utils, only: clean_iodesc_list
+    use dyn_tests_utils, only: vc_height
+    use dynconst, only: constant_g => gravit, constant_pi => pi
+    use inic_analytic, only: analytic_ic_active, dyn_set_inic_col
+    use physconst, only: constant_cpd => cpair, constant_p0 => pref, constant_rd => rair, constant_rv => rh2o
```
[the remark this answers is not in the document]
kuanchihwang: I have switched to only use constants from `dynconst`. This should be resolved by the following two commits: https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/b7dccf3e5e708c46edd7062d58396bcdc5ef59fe https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/03cf4e001348d372c7362d2db6162aafbd6d00b4

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +175,575 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+        integer :: ierr
+        logical, allocatable :: is_water_species(:)
+        type(file_desc_t), pointer :: pio_init_file => null()
+        type(file_desc_t), pointer :: pio_topo_file => null()
+
+        allocate(constituent_name(num_advected), stat=ierr)
+        call check_allocate(ierr, 'dyn_init', 'constituent_name', 'dyn_comp', __LINE__)
+
+        allocate(is_water_species(num_advected), stat=ierr)
+        call check_allocate(ierr, 'dyn_init', 'is_water_species', 'dyn_comp', __LINE__)
```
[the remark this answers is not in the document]
kuanchihwang: This should be resolved by the commit https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/ef3f0761d6714e5328c99f39b8af48c59325f169. I included array shapes in all calls to `check_allocate` in `dyn_grid` as well.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +175,575 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+            else
+                ! Namelist option that controls if constituents are to be read from the file.
+                if (readtrace) then
+                    ! Read variables that belong to the "input" stream in MPAS.
+                    call mpas_dynamical_core % read_write_stream(pio_init_file, 'r', 'input')
+
+                    ! TODO:
+                    ! `cnst_init_default` or its replacement is not yet implemented in CAM-SIMA.
+                    ! Default constituent initialization should be performed here
+                    ! for constituents that fail to be read from the file.
```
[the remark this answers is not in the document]
kuanchihwang: Thanks for the heads-up! I have implemented default constituent initialization for MPAS dycore in the commit https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/d3567bf2793e631053314022dca16613ba83dfd7.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +175,575 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+                else
+                    ! Read variables that belong to the "input" stream in MPAS, excluding constituents.
+                    call mpas_dynamical_core % read_write_stream(pio_init_file, 'r', 'input-scalars')
+
+                    ! TODO:
+                    ! `cnst_init_default` or its replacement is not yet implemented in CAM-SIMA.
+                    ! Default constituent initialization should be performed here
+                    ! because constituents are not read from the file.
```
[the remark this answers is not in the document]
kuanchihwang: As above, this should be resolved by the commit https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/d3567bf2793e631053314022dca16613ba83dfd7.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +175,575 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+            allocate(p_sfc(ncells_solve), stat=ierr)
+            call check_allocate(ierr, 'set_analytic_initial_condition', 'p_sfc', 'dyn_comp', __LINE__)
+
+            buffer_1d_real(:) = 0.0_kind_r8
+
+            call dyn_set_inic_col(vc_height, lat_rad, lon_rad, global_grid_index, zint=z_int, ps=buffer_1d_real)
+
+            p_sfc(:) = buffer_1d_real(:)
+
+            deallocate(buffer_1d_real)
```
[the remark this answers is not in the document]
kuanchihwang: Of course. I have avoided the unnecessary array allocation and copy in the commit https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/7a6451c936d57dbcaea05cd504a26614b7787bad.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -894,7 +1225,9 @@ subroutine dyn_mpas_init_stream_with_pool(self, mpas_pool, mpas_stream, pio_file
 
         character(*), parameter :: subname = 'dyn_mpas_subdriver::dyn_mpas_init_stream_with_pool'
         character(strkind) :: stream_filename
-        integer :: i, ierr, ndims, stream_format, varid
+        integer :: i, ierr, stream_format
+        logical, allocatable :: var_is_present(:)
+        logical, allocatable :: var_is_tkr_compatible(:)
```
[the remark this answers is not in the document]
kuanchihwang: "TKR compatible" refers to "type, kind and rank compatible". It is a terminology that is used in the Fortran standard (e.g., "15.4.3.4.5 Restrictions on generic declarations"). I actually had code comments about it [here](https://github.com/ESCOMP/CAM-SIMA/blob/724d5c859556d4ed9bc13de039b574f0810b7252/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90#L1818).

I have added more code comments for better clarity in the commit https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/a87368ac1125ca20e2ce73bbf6ed7c6a030afefc.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -1408,114 +1805,515 @@ pure function index_unique(array)
...
+                                call self % model_error('Failed to allocate var_name_list', subname, __LINE__)
+                            end if
+
+                            var_name_list(:) = field_1d_char % constituentnames(:)
+                        end if
+
+                        nullify(field_1d_char)
+                    case default
+                        call self % model_error('Unsupported variable rank ' // stringify([var_info % rank]), &
+                            subname, __LINE__)
```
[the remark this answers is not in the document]
kuanchihwang: Yes, I retrospectively changed all similar error messages to include variable name in the commit https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/b3f1937bb868856590c9380d97c5897ee6600ae0.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -1408,114 +1805,515 @@ pure function index_unique(array)
...
+                            var_name_list(:) = field_5d_real % constituentnames(:)
+                        end if
+
+                        nullify(field_5d_real)
+                    case default
+                        call self % model_error('Unsupported variable rank ' // stringify([var_info % rank]), &
+                            subname, __LINE__)
+                end select
+            case default
+                call self % model_error('Unsupported variable type "' // trim(adjustl(var_info % type)) // &
```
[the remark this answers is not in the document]
kuanchihwang: As above, this should be resolved by https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/b3f1937bb868856590c9380d97c5897ee6600ae0.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -1264,70 +1590,141 @@ subroutine add_stream_attribute_1d(attribute_name, attribute_value)
...
+                select case (stream_name_operator)
+                    case ('+')
+                        var_info_list = [var_info_list, var_info_list_buffer]
+                    case ('-')
+                        do j = 1, size(var_info_list_buffer)
+                            var_name_list = var_info_list % name
+                            var_info_list = pack(var_info_list, var_name_list /= var_info_list_buffer(j) % name)
+                        end do
+                    case default
+                        ! Do nothing for unknown operators. Should not happen at all.
```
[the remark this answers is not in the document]
kuanchihwang: The quoted code section here is inside a **pure** function, `parse_stream_name`. We cannot have I/O statements in pure procedures.

The debuggability should still be fine though. This function returns a list (rank-1 array) of `var_info_type`. Wherever it is called, there are debug messages below it like [here](https://github.com/ESCOMP/CAM-SIMA/blob/724d5c859556d4ed9bc13de039b574f0810b7252/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90#L1294-L1305), which prints the elements of that list. A bad operator will cause the resulting list to either contain extra or missing elements. Therefore, by inspecting debug messages, this should be fairly obvious to spot.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +175,575 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+            call mpas_dynamical_core % exchange_halo('rho')
+            call mpas_dynamical_core % exchange_halo('theta')
+        end subroutine set_mpas_state_rho_theta
+
+        !> Set MPAS state `rho_base` (i.e., base state dry air density) and `theta_base` (i.e., base state potential temperature).
+        !> (KCW, 2024-05-21)
+        subroutine set_mpas_state_rho_base_theta_base()
+            integer :: i, k
+            integer :: ierr
+            real(kind_r8), parameter :: t_base = 250.0_kind_r8 ! Base state temperature (K) of dry isothermal atmosphere.
```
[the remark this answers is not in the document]
kuanchihwang: For MPAS, initial conditions are generated by an independent program (`init_atmosphere`), which is separate from the model itself (`atmosphere_model`). The dycore component, which CAM-SIMA uses, resides in the latter, and we sparse checkout only the necessary source code to build it.

In addition, this constant is defined in the former like [here](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.2.0/src/core_init_atmosphere/mpas_init_atm_cases.F#L2649), and it is also not a module variable that we can use. Therefore, I think that redefining it here is necessary.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +175,575 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+            call mpas_dynamical_core % get_variable_pointer(theta_base, 'diag', 'theta_base')
+            call mpas_dynamical_core % get_variable_pointer(zz, 'mesh', 'zz')
+
+            ! Set `rho_base` and `theta_base` column by column. This way, peak memory usage can be reduced.
+            do i = 1, ncells_solve
+                do k = 1, pver
+                    ! Derive `p_base` by hypsometric equation.
+                    ! The formulation used here is exact and identical to MPAS.
+                    p_base(k) = constant_p0 * exp(-0.5_kind_r8 * (zgrid(k, i) + zgrid(k + 1, i)) / &
+                        (constant_rd * t_base / constant_g))
```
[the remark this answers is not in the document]
kuanchihwang: The above rationale holds here, too. The hypsometric equation is found [here](https://github.com/MPAS-Dev/MPAS-Model/blob/v8.2.0/src/core_init_atmosphere/mpas_init_atm_cases.F#L4862-L4863). It is also inlined so that we cannot use it.

However, thanks to your suggestion, I can indeed factor out the equations used here. This commit https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/77cf091c24a89ee229b2ba291f928b85cb9997d5 should hopefully resolve this.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +175,575 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+        allocate(constituent_name(num_advected), stat=ierr)
+        call check_allocate(ierr, 'dyn_init', 'constituent_name', 'dyn_comp', __LINE__)
+
+        allocate(is_water_species(num_advected), stat=ierr)
+        call check_allocate(ierr, 'dyn_init', 'is_water_species', 'dyn_comp', __LINE__)
+
+        do i = 1, num_advected
+            constituent_name(i) = const_name(i)
+            is_water_species(i) = const_is_water_species(i)
+        end do
```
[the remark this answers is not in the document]
kuanchihwang: The two variables are used to call `dyn_mpas_define_scalar` [[1]](https://github.com/ESCOMP/CAM-SIMA/blob/724d5c859556d4ed9bc13de039b574f0810b7252/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90#L855), which informs MPAS, at run-time, about **all** constituent names and their corresponding waterness.

Indeed, I can choose to pass `const_name` and `const_is_water_species` as procedure pointers and call them from MPAS side. But, this will involve assumptions about their procedure interfaces, which is not ideal.

I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/724d5c859556d4ed9bc13de039b574f0810b7252 to directly use them elsewhere.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -32,15 +51,35 @@ module dyn_comp
 
     public :: dyn_debug_print
     public :: mpas_dynamical_core
+    public :: ncells, ncells_solve, nedges, nedges_solve, nvertices, nvertices_solve, nvertlevels
+    public :: ncells_global, nedges_global, nvertices_global, ncells_max, nedges_max
+    public :: sphere_radius
+    public :: deg_to_rad, rad_to_deg
```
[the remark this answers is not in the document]
kuanchihwang: I totally agree that this is really suboptimal, but the reason is explained in the commit message of https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/932ef82f6c3d3a283e214b9894facb333f61ed62.

The instance of MPAS dynamical core resides in `dyn_comp`. `dyn_grid` depends on `dyn_comp` because it needs to interact with the instance. So, if `dyn_comp` were to depend on `dyn_grid` for grid variables, there would be a circular dependency issue.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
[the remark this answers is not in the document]
nusbaume: Ahh, yeah, I missed that it was in a pure function.  I agree that doing the debug-related checks after the function has returned is fine.  

If we ever find that this is a common problem (which I doubt it will be) then we can always have it return very specific quantities in `var_info_type` that is then checked immediately afterwards, but I certainly wouldn't worry about it for this PR (if ever).

src/dynamics/mpas/dyn_comp.F90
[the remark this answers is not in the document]
nusbaume: Thanks for the explanation!  In that case I think this is fine as-is.

src/dynamics/mpas/dyn_comp.F90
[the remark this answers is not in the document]
nusbaume: Thanks for the explanation!  If this constant is in an entirely different section of the MPAS system then I agree that just re-defining it here for now is fine.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -620,22 +618,37 @@ subroutine set_mpas_state_rho_theta()
 
                 ! Piecewise integrate hypsometric equation to derive `p_mid_col(1)`.
                 ! The formulation used here is exact.
-                p_mid_col(1) = p_sfc(i) * &
-                    exp(-0.5_kind_r8 * (zgrid(2, i) - zgrid(1, i)) * &
-                    constant_g / (constant_rd * tm_mid_col(1)) * (1.0_kind_r8 + qv_mid_col(1)))
+                ! p_mid_col(1) = p_sfc(i) * &
+                !     exp(-0.5_kind_r8 * (zgrid(2, i) - zgrid(1, i)) * &
+                !     constant_g / (constant_rd * tm_mid_col(1)) * (1.0_kind_r8 + qv_mid_col(1)))
```
nusbaume, line 623: I might remove the commented-out code here.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -620,22 +618,37 @@ subroutine set_mpas_state_rho_theta()
...
-                    p_mid_col(k) = p_mid_col(k - 1) * &
-                        exp(-0.5_kind_r8 * (zgrid(k    , i) - zgrid(k - 1, i)) * &
-                        constant_g / (constant_rd * tm_mid_col(k - 1)) * (1.0_kind_r8 + qv_mid_col(k - 1))) * &
-                        exp(-0.5_kind_r8 * (zgrid(k + 1, i) - zgrid(k    , i)) * &
-                        constant_g / (constant_rd * tm_mid_col(k    )) * (1.0_kind_r8 + qv_mid_col(k    )))
+                    ! p_mid_col(k) = p_mid_col(k - 1) * &
+                    !     exp(-0.5_kind_r8 * (zgrid(k    , i) - zgrid(k - 1, i)) * &
+                    !     constant_g / (constant_rd * tm_mid_col(k - 1)) * (1.0_kind_r8 + qv_mid_col(k - 1))) * &
+                    !     exp(-0.5_kind_r8 * (zgrid(k + 1, i) - zgrid(k    , i)) * &
+                    !     constant_g / (constant_rd * tm_mid_col(k    )) * (1.0_kind_r8 + qv_mid_col(k    )))
```
nusbaume, line 637: I might remove the commented-out code here.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -620,22 +618,37 @@ subroutine set_mpas_state_rho_theta()
...
+                            tm_mid_col(k - 1) / (1.0_kind_r8 + qv_mid_col(k - 1)), &
+                            zgrid(k, i)), &
+                        zgrid(k, i), &
+                        tm_mid_col(k) / (1.0_kind_r8 + qv_mid_col(k)), &
+                        0.5_kind_r8 * (zgrid(k + 1, i) + zgrid(k, i)))
                 end do
 
                 rho(:, i) = p_mid_col(:) / (constant_rd * tm_mid_col(:))
-                theta(:, i) = t_mid(:, i) * ((constant_p0 / p_mid_col(:)) ** (constant_rd / constant_cpd))
+                ! theta(:, i) = t_mid(:, i) * ((constant_p0 / p_mid_col(:)) ** (constant_rd / constant_cpd))
```
nusbaume, line 650: I might remove the commented-out code here.

nusbaume (approved):
Looks good to me now!  I did have some final cleanup requests, but it doesn't need a re-review (from me at least).  Thanks!

src/dynamics/mpas/dyn_comp.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/3ca360bb44c2935ae58afce307c82a2829cc51b6 to remove the commented code.

src/dynamics/mpas/dyn_comp.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/3ca360bb44c2935ae58afce307c82a2829cc51b6 to remove the commented code.

src/dynamics/mpas/dyn_comp.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/3ca360bb44c2935ae58afce307c82a2829cc51b6 to remove the commented code.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -18,9 +18,9 @@ module dyn_comp
     use cam_logfile, only: debug_output, debugout_none, iulog
     use cam_pio_utils, only: clean_iodesc_list
     use dyn_tests_utils, only: vc_height
-    use dynconst, only: constant_g => gravit, constant_pi => pi
+    use dynconst, only: constant_cpd => cpair, constant_g => gravit, constant_p0 => pref, constant_pi => pi, &
+                        constant_rd => rair, constant_rv => rh2o
```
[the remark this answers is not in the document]
kuanchihwang: Yes, I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/9cc38a476cc6b27bd50d4123e225bb954bfa5112 to move them to `dynconst`.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -703,37 +740,50 @@ end subroutine dyn_mpas_init_phase2
...
+    subroutine dyn_mpas_init_phase3(self, number_of_constituents, pio_file)
+        class(mpas_dynamical_core_type), intent(inout) :: self
+        integer, intent(in) :: number_of_constituents
         type(file_desc_t), pointer, intent(in) :: pio_file
 
         character(*), parameter :: subname = 'dyn_mpas_subdriver::dyn_mpas_init_phase3'
         character(strkind) :: mesh_filename
         integer :: mesh_format
+        integer, pointer :: num_scalars => null()
+        type(mpas_pool_type), pointer :: mpas_pool => null()
```
mgduda, line 767: Rather than initialize these here (adding the implicit `SAVE` attribute), is there a reason to not just nullify these either at the beginning of this routine or just before they are (hopefully) set? Would doing this remove the need to explicitly nullify these at the end of the subroutine?

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -703,37 +740,50 @@ end subroutine dyn_mpas_init_phase2
...
 
-        call self % debug_print('Number of constituents is ', [cam_pcnst])
+        ! In MPAS, there must be at least one constituent, `qv`, which denotes water vapor mixing ratio.
+        ! Because MPAS has some hard-coded array accesses through the `index_qv` index, it will crash
+        ! (i.e., segmentation fault due to invalid memory access) if `qv` is not allocated.
+        if (number_of_constituents < 1) then
+            self % number_of_constituents = 1
+        else
+            self % number_of_constituents = number_of_constituents
+        end if
```
mgduda, line 778: Perhaps it's less clear, but just for consideration since it's more compact, it seems like the following would also work:
```fortran
self % number_of_constituents = max(1, number_of_constitutents)
```

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -759,9 +809,290 @@ subroutine dyn_mpas_init_phase3(self, cam_pcnst, pio_file)
...
+            'water_vapor_mixing_ratio_wrt_moist_air', &
+            'water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water' &
+        ]
+
+        character(*), parameter :: subname = 'dyn_mpas_subdriver::dyn_mpas_define_scalar'
+        integer :: i, j, ierr
+        integer :: index_qv, index_water_start, index_water_end
+        integer :: time_level
+        type(field3dreal), pointer :: field_3d_real => null()
+        type(mpas_pool_type), pointer :: mpas_pool => null()
```
mgduda, line 874: Same question here as earlier: should be just `nullify` these within the subroutine rather than implicitly saving them with initialization?

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -759,9 +809,290 @@ subroutine dyn_mpas_init_phase3(self, cam_pcnst, pio_file)
...
+
+            self % constituent_name(:) = adjustl(constituent_name)
+
+            allocate(self % is_water_species(self % number_of_constituents), stat=ierr)
+
+            if (ierr /= 0) then
+                call self % model_error('Failed to allocate is_water_species', subname, __LINE__)
+            end if
+
+            self % is_water_species(:) = is_water_species
```
mgduda, line 928: In my opinion, it makes the code a bit clearer to add `(:)` to `is_water_species`; i.e.,
```fortran
self % is_water_species(:) = is_water_species(:)
```

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -32,15 +54,31 @@ module dyn_comp
...
     public :: dyn_debug_print
     public :: mpas_dynamical_core
+    public :: ncells, ncells_solve, nedges, nedges_solve, nvertices, nvertices_solve, nvertlevels
+    public :: ncells_global, nedges_global, nvertices_global, ncells_max, nedges_max
+    public :: sphere_radius
 
+    !> NOTE:
+    !> This derived type is not used by MPAS dynamical core. It exists only as a placeholder because CAM-SIMA requires it.
+    !> Developers/Maintainers/Users who wish to interact with MPAS dynamical core may do so by using the "instance/object"
+    !> below.
```
mgduda, line 64: A question about planned future development stemming this comment: do we not plan on using `dyn_import_t` and `dyn_export_t` ever when it comes to physics-dynamics coupling?

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +174,640 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+            surface_geopotential(:) = 0.0_kind_r8
+            surface_geometric_height(:) = 0.0_kind_r8
+
+            call cam_read_field('PHIS', pio_file, surface_geopotential, success, &
+                gridname='cam_cell', timelevel=1, log_output=(debug_output > debugout_none))
+
+            if (success) then
+                surface_geometric_height(:) = surface_geopotential(:) / constant_g
+            else
+               call endrun('Failed to find variable "PHIS"', subname, __LINE__)
```
mgduda, line 324: It's really inconsequential, but should there be an extra space of indentation here?

Also, to my eye it's a bit more compact and intuitive to write this block as:
```fortran
if (.not. success) then
   call endrun('Failed to find variable "PHIS"', subname, __LINE__)
end if

surface_geometric_height(:) = surface_geopotential(:) / constant_g
```
But I offer this just as a suggestion for consideration.

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +174,640 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+
+        if (associated(pio_file)) then
+            call dyn_debug_print('Topography file is used')
+
+            if (.not. pio_file_is_open(pio_file)) then
+                call endrun('Invalid PIO file descriptor', subname, __LINE__)
+            end if
+
+            allocate(surface_geopotential(ncells_solve), stat=ierr)
+            call check_allocate(ierr, 'check_topography_data', 'surface_geopotential(ncells_solve)', 'dyn_comp', __LINE__)
```
mgduda, line 310: Rather than `'check_topography_data'` as the second argument, would it be better to use `subname` instead?

src/dynamics/mpas/dyn_comp.F90
```diff
@@ -139,13 +174,640 @@ pure function sec_to_hour_min_sec(sec) result(hour_min_sec)
...
+        subroutine init_shared_variable()
+            integer :: ierr
+            integer :: k
+            integer, pointer :: indextocellid(:) => null()
+            real(kind_r8), pointer :: lat_deg(:) => null(), lon_deg(:) => null()
+
+            call dyn_debug_print('Preparing to set analytic initial condition')
+
+            allocate(global_grid_index(ncells_solve), stat=ierr)
+            call check_allocate(ierr, 'set_analytic_initial_condition', 'global_grid_index(ncells_solve)', 'dyn_comp', __LINE__)
```
mgduda, line 384: Would it be worth defining a `subname` parameter for `'set_analytic_conditions::init_shared_variables'` to use in calls to `check_allocate` throughout this routine?

mgduda (changes requested):
I haven't yet had a chance to try different types of ICs, so I'll likely make a second pass with some further comments.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -1573,6 +2370,611 @@ subroutine dyn_mpas_exchange_halo(self, field_name)
...
+        integer, pointer :: nvertlevels => null(), maxedges => null(), maxedges2 => null(), num_scalars => null()
+        logical, pointer :: config_do_restart => null()
+        real(rkind), pointer :: config_dt => null()
+        type(field0dreal), pointer :: field_0d_real => null()
+        type(mpas_pool_type), pointer :: mpas_pool => null()
+        type(mpas_time_type) :: mpas_time
+
+        call self % debug_print(subname // ' entered')
+
+        if (real(coupling_time_interval, rkind) < 1.0_rkind) then
```
mgduda, line 2568: Could this check be simplified? If `coupling_time_interval` is an integer, `real(coupling_time_interval, rkind) < 1.0_rkind` iff `coupling_time_interval` is zero or a negative integer, right? Would the following work?
```fortran
if (coupling_time_interval < 1) then
```

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -1573,6 +2370,611 @@ subroutine dyn_mpas_exchange_halo(self, field_name)
...
+        pure function almost_divisible(a, b)
+            real(rkind), intent(in) :: a, b
+            logical :: almost_divisible
+
+            real(rkind) :: error_tolerance
+
+            error_tolerance = epsilon(1.0_rkind) * max(abs(a), abs(b))
+
+            if (almost_equal(mod(abs(a), abs(b)), 0.0_rkind, absolute_tolerance=error_tolerance) .or. &
+                almost_equal(mod(abs(a), abs(b)), abs(b), absolute_tolerance=error_tolerance)) then
```
mgduda, line 2748: Could you clarify why this second `almost_equal` is needed?

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -1573,6 +2370,611 @@ subroutine dyn_mpas_exchange_halo(self, field_name)
...
+        call self % debug_print(subname // ' entered')
+
+        if (real(coupling_time_interval, rkind) < 1.0_rkind) then
+            call self % model_error('Invalid coupling time interval ' // stringify([real(coupling_time_interval, rkind)]), &
+                subname, __LINE__)
+        end if
+
+        call self % get_variable_pointer(config_dt, 'cfg', 'config_dt')
+
+        if (config_dt < 1.0_rkind) then
```
mgduda, line 2575: I wouldn't say that a sub-one second timestep is necessarily an invalid timestep for the MPAS-A dycore. Admittedly it's unlikely that we'll be using such small timesteps any time soon in CAM-SIMA. But, it does seem like a somewhat arbitrary choice to disallow, say, a 0.5 s timestep, but to allow a 1.5 s timestep.
Perhaps instead we could simply check for negative values of `config_dt`?

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
mgduda: Ok, stepping away from code for a bit can do wonders. If `a` is slightly less an an integer multiple of `b` we need this check.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/155b51c4e3f01e2b6114979d2c3051ff7cb27f55 to resolve this.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/b2ef93e070812e3b2f9a42ec779865762985e97b to resolve this. Other similar instances in the code also follow this style.

src/dynamics/mpas/dyn_comp.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/03768738e0dbcc08112a8cdca9682ed59e7d18fe to make the conditional more intuitive.

src/dynamics/mpas/dyn_comp.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/149cbca853d77f7ba5dbc64ac4c14dc5c2b9d8e0 to resolve this. I did not pass `subname` directly to `check_allocate` because it contains duplicate information with the 4th argument, `file`, of `check_allocate`.

src/dynamics/mpas/dyn_comp.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/d18264ba2467694d053246789215e23c2048f176 to also add `subname` to internal subroutines.

Note that with this commit, `fortls` will complain about variable shadowing ("Variable "subname" masks variable in parent scope"). It should be fine because that is just a diagnostic.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/a731e9769498b5a3b374c2329aaf97eaec2ff853 to avoid explicit initialization of pointers.

Thanks for catching this. I don't know why I thought the rule doesn't apply to pointers.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
kuanchihwang: Same as the above. I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/a731e9769498b5a3b374c2329aaf97eaec2ff853 to resolve this.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -11,16 +11,25 @@ module dyn_mpas_subdriver
 
     ! Modules from external libraries.
     use mpi, only: mpi_comm_null, mpi_comm_rank, mpi_success
+    use netcdf, only: nf90_char, nf90_int, nf90_float, nf90_double
```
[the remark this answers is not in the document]
kuanchihwang: Yes, I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/3308a26ac86d5162b0369cecca487bef6454dfa1 to resolve this.

The reason that I used type constants from `netcdf` is due to the following.

`pio_inq_vartype` is a generic interface [[1]](https://github.com/NCAR/ParallelIO/blob/pio2_6_2/src/flib/pio_nf.F90#L247). Eventually, the `inq_vartype_id` function [[2]](https://github.com/NCAR/ParallelIO/blob/pio2_6_2/src/flib/pio_nf.F90#L1333) is called, and then the `PIOc_inq_vartype` C function [[3]](https://github.com/NCAR/ParallelIO/blob/pio2_6_2/src/clib/pio_nc.c#L1031) is called through ISO C binding.

`PIOc_inq_vartype` just calls the `PIOc_inq_var` function [[4]](https://github.com/NCAR/ParallelIO/blob/pio2_6_2/src/clib/pio_nc.c#L848), and finally, the `nc_inq_var` function [[5]](https://github.com/NCAR/ParallelIO/blob/pio2_6_2/src/clib/pio_nc.c#L932) from NetCDF is used to actually determine the variable type.

The PIO type constants in Fortran are defined separately [[6]](https://github.com/NCAR/ParallelIO/blob/pio2_6_2/src/flib/pio_types.F90#L133). Therefore, we can come to the conclusion that the type constants returned by `pio_inq_vartype` are entirely determined by NetCDF, and have nothing to do with the PIO type constants.

If PIO type constants were to become out-of-sync with NetCDF type constants, it would be difficult to debug.

src/dynamics/mpas/dyn_comp.F90
kuanchihwang: From what I understand, the `dyn_import_t` and `dyn_export_t` derived types are used like a container of pointers to dynamical state variables.

This functionality is no longer needed because we now have a more generic facility: `get_variable_pointer` [[1]](https://github.com/ESCOMP/CAM-SIMA/blob/d18264ba2467694d053246789215e23c2048f176/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90#L3139), which allows for drawing the reference of any internal variables from MPAS dynamical core.

Therefore, when implementing dynamics-physics coupling, we can just draw pointers to those necessary internal variables from MPAS dynamical core by calling `get_variable_pointer`.

kuanchihwang: If you would like to preview how it works, you can look at the `implement-dyn-run` [[2]](https://github.com/kuanchihwang/CAM-SIMA/blob/develop/implement-dyn-run/src/dynamics/mpas/dyn_coupling.F90) branch of my fork, where I am currently implementing the time integration + dynamics-physics coupling of MPAS dynamical core in CAM-SIMA.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/6428560b42e2eee6e8db89d1ffe04f1016212cd9 to simplify this check.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
kuanchihwang: I pushed https://github.com/ESCOMP/CAM-SIMA/pull/267/commits/6428560b42e2eee6e8db89d1ffe04f1016212cd9 to relax this check. I think zero should also be invalid because it would make the time integration stall.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
mgduda: It's possible that I'm wrong, but `5.1 Type declaration statements" of the 2003 standard states:
```
33 If initialization is =>null-init, object-name shall be a pointer, and its initial association status is disas-
34 sociated.
35 The presence of initialization implies that object-name is saved, except for an object-name in a named
36 common block or an object-name with the PARAMETER attribute. The implied SAVE attribute may
37 be reaffirmed by explicit use of the SAVE attribute in the type declaration statement, by inclusion of
38 the object-name in a SAVE statement (5.2.12), or by the appearance of a SAVE statement without a
39 saved-entity -list in the same scoping unit.
```

src/dynamics/mpas/dyn_comp.F90
mgduda: Seems reasonable enough to me!

mgduda (approved):
Apologies for taking so long to complete this review! This all seems good to me.

PR #269, opened 2024-06-12
Simplify snapshot checking reporting and functionality

### LINKED ISSUE #268: Need to simplify snapshot (ncdata_check) testing
### Issue Type

Infrastructure Update

### Issue Description

The current implementation of using CAM snapshot files to test CCPP-ized physics schemes in CAM-SIMA is overly complex due to technical debt from the beginning development period of the model, and has started to cause confusion among several CAM-SIMA and CPPP physics developers.  Thus it would be good to simplify the implementation of this testing mechanism by doing the following:

1.  Use the actual  current CAM-SIMA time step as the time slice/data frame used for the snapshot files, instead of the model time step + 2.

2.  Only perform snapshot comparisons during the CAM-SIMA run phase, instead of during all model phases (e.g. initialize, run, finalize).

Along with those two modifications, it would also help to have CAM-SIMA add a line to the log file whenever the time step has advanced, so that users and developers can know on what time step the model is actually on at any given point in the log output.
 

### Will this change answers?

No

### Will you be implementing this yourself?

Yes

nusbaume:
This PR ensures that snapshot checks (e.g. `ncdata_check`) are only run during actual model run steps (not initialize or finalize steps), and that there is no offset between the model timestep and the input or check file time.

This PR also adds a new `current_timestep_number` registry variable which is updated every time the model time step advances, as well as a new log print statement that indicates which model timestep CAM-SIMA is currently on.

Fixes #268 

## Tests run:

Ran a physics testbed configuration with both Kessler and Held-Suarez physics for both a single timestep and for multiple timesteps (no differences found for any of the tests).

Also ran the official regression build tests on derecho for both GNU and Intel.

### DIFF
--- a/src/physics/utils/phys_comp.F90
+++ b/src/physics/utils/phys_comp.F90
@@ -185,11 +185,9 @@ CONTAINS
       ! Physics needs to read in all data not read in by the dycore
       ncdata => initial_file_get_id()
 
-      ! data_frame is the next input frame for physics input fields
-      ! Frame 1 is skipped for snapshot files
-      !!XXgoldyXX: This section needs to have better logic once we know if
-      !!           this is a physics test bench run.
-      data_frame = get_nstep() + 2
+      ! data_frame is the next input frame for
+      ! physics fields that must be read from a file:
+      data_frame = get_nstep()
 
       ! Initialize host model variables that must be done each time step:
       call physics_types_tstep_init()
@@ -237,15 +235,18 @@ CONTAINS
 
    end subroutine phys_run2
 
-   subroutine phys_timestep_final()
+   subroutine phys_timestep_final(do_ncdata_check)
       use time_manager,   only: get_nstep
       use cam_abortutils, only: endrun
       use cam_initfiles,  only: unset_path_str
       use cam_ccpp_cap,   only: cam_ccpp_physics_timestep_final
       use physics_inputs, only: physics_check_data
 
+      ! Subroutine inputs
+      logical, intent(in) :: do_ncdata_check
+
       ! Local variables
-      integer                              :: data_frame
+      integer             :: data_frame
 
       ! Finalize the time step
       call cam_ccpp_physics_timestep_final(phys_suite_name)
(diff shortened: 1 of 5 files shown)


cacraigucar (approved):
Not sure I totally understand all the logic in here, but I don't see any problems in the recently added code.

src/physics/utils/phys_comp.F90
```diff
@@ -237,32 +235,35 @@ subroutine phys_run2()
...
-      data_frame = get_nstep() + 2
+      ! data_frame is the next input frame for
+      ! physics snapshot validation fields
+      data_frame = get_nstep()
 
       ! Determine if physics_check should be run:
       if (trim(ncdata_check) /= trim(unset_path_str)) then
-         call physics_check_data(ncdata_check, suite_names, data_frame,       &
-            min_difference, min_relative_value)
+         if (do_ncdata_check) then
```
peverwhee, line 263: do we still want to include the check if ncdata_check is populated in order to proceed to the check? I see that there's logic in physics_check data to issue a warning ('WARNING: Namelist variable ncdata_check is UNSET. Model will run, but physics check data will not be printed') if it's not, but I can also imagine that warning would be confusing if you had no intention of running the check.

peverwhee (commented):
teeny tiny question

src/physics/utils/phys_comp.F90
nusbaume: That's a good question!  I guess my opinion would be to just skip the call to `physics_check_data` entirely if we can avoid it, so I would prefer to leave this `ncdata_check` if-statement here as-is.

That being said, I am also not sure if the extra internal logic inside of `physics_check_data` you mentioned is really hurting anything either (beyond maybe a few wasted clock cycles).  So that combined with the fact that I am not totally sure what the long-term plan for the `physics_check_data` routine will be (e.g. will it eventually be called elsewhere in SIMA?  Removed entirely?) makes me lean towards just leaving everything as-is on that front for now as well.  Of course if you or @cacraigucar feel strongly one way or the other please let me know.  Thanks!

cacraigucar: I like avoiding writing out the WARNING message for runs where we are not performing checks (normal CAM-SIMA runs in the future?)  I figure when we get to that stage, we can make sure that everything is running "best" for CAM-SIMA and make any required tweeks at that time.  So I don't have a strong feeling one way or the other.

peverwhee: Well, turns out I can't read. The `if (trim(ncdata_check) /= trim(unset_path_str)) then` is still there and I thought it was gone. So.. agreed and nevermind!

peverwhee (approved):

PR #272, opened 2024-07-03
Fix constituent handling

peverwhee:
**Main fix:**

* physics/utils/physics_data.F90: update the logic to not return after determining if a variable is a constituent. This keeps the "initialized" logic so that SIMA doesn't try to read in constituents that are initialized by the SE dycore. 
  * Basically, only set find_input_name_idx to "const_idx" (the parameter that determines if a variable is a constituent to be read in from file) if it's not marked as initialized or protected.

**Smaller fixes**
* dynamics/se/namelist_definition_se_dycore.xml: add namelist default for ne3 grid
* data/write_init_files.py: trim constituent standard names output to log


src/physics/utils/physics_data.F90
```diff
@@ -56,21 +56,20 @@ integer function find_input_name_idx(stdname, use_init_variables, constituent_in
...
       find_input_name_idx = no_exist_idx
       constituent_index = no_exist_idx
+      is_constituent = .false.
 
       !First check if quantity is a constituent:
       call const_get_index(trim(stdname), find_input_name_idx, abort=.false., warning=.false.)
       if (find_input_name_idx < 0) then
          find_input_name_idx = no_exist_idx
       else
          constituent_index = find_input_name_idx
```
nusbaume, line 71: While we are here, could we simplify this code block to the following, given that `find_inpute_name_idx` is already initialized to `no_exist_idx`:

```suggestion
      if (find_input_name_idx >= 0) then
         constituent_index = find_input_name_idx
```

nusbaume (approved):
Thanks for tracking down this bug!  I just had one (optional) code clean-up request.

Also, an FYI that I have added instructions for the current CAM-SIMA regression tests here:
https://github.com/ESCOMP/CAM-SIMA/wiki/CAM-SIMA-testing#regression-testing-manual

Please let me know if you have any questions!

src/physics/utils/physics_data.F90
peverwhee: ah good call! fixed!

peverwhee:
tests run and passed through model_build:
SMS_Ln2.ne3pg3_ne3pg3_mg37.FPHYStest.derecho_intel.cam-outfrq_kessler_derecho_nooutput
SMS_Ln2.mpasa480_mpasa480.FKESSLER.derecho_gnu.cam-outfrq_kessler_mpas_derecho_nooutput
SMS_Ln2.ne3pg3_ne3pg3_mg37.FPHYStest.derecho_gnu.cam-outfrq_held_suarez_derecho_nooutput

mwaxmonsky (approved):
Looks great @peverwhee!

peverwhee:
Requesting re-review for new fix:

* control/cam_comp.F90: move call to stepon_run1 to cam_timestep_init so that host variables are initialized before any call to CCPP physics. Also rename stepon_run1 to stepon_timestep_init
* dynamics/*/stepon.F90: rename stepon_run1 to stepon_timestep_init

All regression tests pass.

nusbaume (approved):

mwaxmonsky (approved):

PR #281, opened 2024-07-25
Add git-fleximod ci test

peverwhee:
Updates git-fleximod to v8.4 and adds the git-fleximod github workflow.

List any changes to the defaults for the boundary datasets: N/A

Describe any substantial timing or memory changes: N/A

Code reviewed by: cacraigucar, mwaxmonsky, nusbaume

List all existing files that have been modified, old files eliminated, new files added and describe the changes:

A .github/workflows/fleximod_test.yaml

derecho/intel/aux_cam: all tests pass
derecho/gnu/aux_cam:   all tests pass

Summarize any changes to answers, i.e.,

what code configurations: N/A
what platforms/compilers: N/A
nature of change (roundoff; larger than roundoff but same climate; new climate): N/A

### DIFF
--- a/.github/workflows/fleximod_test.yaml
+++ b/.github/workflows/fleximod_test.yaml
@@ -1,0 +1,24 @@
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  fleximod-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # oldest supported and latest supported
+        python-version: ["3.7", "3.x"]
+    steps:
+      - id: checkout-CESM
+        uses: actions/checkout@v4
+      - id: run-fleximod
+        run: |
+          $GITHUB_WORKSPACE/bin/git-fleximod update
+          $GITHUB_WORKSPACE/bin/git-fleximod test
+#      - name: Setup tmate session
+#        if: ${{ failure() }}
+#        uses: mxschmitt/action-tmate@v3
+        
+          
--- a/.github/workflows/fleximod_test.yaml
+++ b/.github/workflows/fleximod_test.yaml
@@ -1,8 +1,6 @@
 on:
-  push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: [ development ]
 jobs:
   fleximod-test:
     runs-on: ubuntu-latest
--- a/.github/workflows/fleximod_test.yaml
+++ b/.github/workflows/fleximod_test.yaml
@@ -1,13 +1,12 @@
 on:
   pull_request:
-    branches: [ development ]
 jobs:
   fleximod-test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         # oldest supported and latest supported
-        python-version: ["3.7", "3.x"]
+        python-version: ["3.8", "3.x"]
     steps:
       - id: checkout-CESM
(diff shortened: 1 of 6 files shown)


.github/workflows/fleximod_test.yaml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - id: checkout-CESM
+        uses: actions/checkout@v4
+      - id: run-fleximod
+        run: |
+          $GITHUB_WORKSPACE/bin/git-fleximod update
+          $GITHUB_WORKSPACE/bin/git-fleximod test
+#      - name: Setup tmate session
+#        if: ${{ failure() }}
+#        uses: mxschmitt/action-tmate@v3
```
mwaxmonsky, line 20: Can these be deleted?  I think we can be copy them from the tmate repo later if needed (https://github.com/mxschmitt/action-tmate?tab=readme-ov-file#getting-started).

mwaxmonsky (approved):
Looks great @peverwhee!  Just one minor comment but nothing that should hold this up.

cacraigucar (approved):
This appears to be identical to what we brought into ESCOMP/CAM, so I am approving this.  I figure if there are any issues, they would show up when it is committed.

.github/workflows/fleximod_test.yaml
cacraigucar: @mwaxmonsky - this is from the official code that Jim released, and is ported to all the components.  I would suggest that we keep it as it is on all the other platforms (since it doesn't sound like it does any harm)

nusbaume (commented):
Should we add some text to the new CAM-SIMA developer documentation that describes (or includes a link to a description of) these tests?  That way we can point developers to something if the tests fail.

peverwhee:
@nusbaume yes! done!

https://escomp.github.io/CAM-SIMA-docs/development/cam-testing/#git-fleximod-tests

.github/workflows/fleximod_test.yaml
```diff
@@ -0,0 +1,22 @@
+on:
+  pull_request:
+    branches: [ development ]
```
nusbaume, line 3: I might have this run on all pull requests instead of just development (as we'll likely only ever do direct commits to `main`):

```suggestion
  pull_request:
```

.github/workflows/fleximod_test.yaml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  fleximod-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # oldest supported and latest supported
+        python-version: ["3.7", "3.x"]
```
nusbaume, line 10: I would bump the minimum version up to `3.8` here:

```suggestion
        python-version: ["3.8", "3.x"]
```

nusbaume (changes requested):
Thanks for adding the documentation!  Just had a couple more suggestions.

.github/workflows/fleximod_test.yaml
peverwhee: comments removed because the workflow has diverged anyway

.github/workflows/fleximod_test.yaml
peverwhee: done!

nusbaume (approved):
Looks good to me now.  Thanks @peverwhee!

Issue #237, opened 2023-11-13
MUSICA linking fails on Derecho


nusbaume:
### What happened?

Currently CAM-SIMA assumes it can link to the `musica` and `jsonfortran` libraries whenever one uses a compset other than `FPHYStest`.  However, if then on tries to build on Derecho it fails with the following error:

```
/usr/bin/ld: cannot find -lmusica
/usr/bin/ld: cannot find -ljsonfortran
collect2: error: ld returned 1 exit status
```

which indicates that the build system is currently unable to find or link to these libraries on Derecho.

Pinging @mattldawson @K20shores @boulderdaze  so that they know about this.

### What are the steps to reproduce the bug?

The following steps on Derecho will reproduce the error, assuming `<CAM-SIMA>` is wherever the CAM-SIMA codebase is located, with externals already checked out:

```
<CAM-SIMA>/cime/scripts/create_newcase --case cam_sima_test --res ne5pg2_ne5pg2_mg37 --mach derecho --compset FKESSLER --compiler gnu --run-unsupported
cd cam_sima_test
./case.setup
./case.build
```

### What CAM-SIMA hash were you using?

7d299b6

### What machine were you running CAM-SIMA on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

GNU

### Path to a case directory, if applicable

/glade/derecho/scratch/nusbaume/model_runs/SIMA_ne5pg2_FKESSLER_test

### Will you be addressing this bug yourself?

No

### Extra info

If anyone does run into this bug, then it can be fixed by simply running:

`./xmlchange CAM_LINKED_LIBS=' '`

In their case directory, and then running `./case.build` again.

K20shores:
Thanks for letting us know. I made an issue on our side so we can track it

mgduda:
Just giving this issue a bump. It would really be nice if we didn't always have to manually `./xmlchange CAM_LINKED_LIBS=...` when setting up cases.

K20shores:
@nusbaume What would be your recommended fix for this? Musica would only be needed when musica chemistry is enabled. It was never the goal to have this always on but I clearly did something wrong.

nusbaume:
@K20shores If I remember correctly the original idea was that we would always link to a MUSICA library even if we weren't actually using it, but I agree that it seems like it might not be worth it at the moment.  @mgduda I believe you can remove this requirement by simply deleting the following line in CAM-SIMA:

[...]

PR #282, opened 2024-08-01
Includes CCPP scheme dependencies in the build

mattldawson:
Uses updated CCPP-generated data to include scheme dependencies in the CAM-SIMA build.

In draft until https://github.com/ESCOMP/atmospheric_physics/pull/102 is merged.


cime_config/config_component.xml
```diff
@@ -307,6 +307,8 @@
     <valid_values></valid_values>
     <default_value> </default_value>
     <values match="last" modifier="additive">
+      <!-- Turn on chemistry if not using physics testbed -->
+      <value compset="^((?!_CAM%PHYSTEST).)*$">-lmusica-fortran -lmusica -lyaml-cpp</value>
```
nusbaume, line 311: We ended up deciding to not include these flags until there was a version of MUSICA we could link to on Derecho, as it has been causing difficulties with other people's workflows.  For a slightly more detailed discussion see issue #237.  

Of course this was assuming that most current CAM-SIMA MUSICA development would be done with the MUSICA dockerfile (where these flags are being set automatically), but if that is not the case just let me know and we can figure out an alternative solution.  Thanks!

nusbaume (changes requested):
Looks good, but did have one change request, which I am happy to discuss if need be.  Thanks!

cime_config/config_component.xml
mattldawson: removed!

nusbaume:
@mattldawson Are you ok with me pushing to this PR?  I found one small issue when running our regression tests.

nusbaume (approved):

ccpp_framework
peverwhee, line 1: Why are ccpp_framework and atmospheric_physics (ncar_ccpp) being added as submodules? I may be behind on some git-fleximod stuff, but I *think* we only need to update the .gitmodules file.

cime_config/config_component.xml
```diff
@@ -307,6 +307,8 @@
     <valid_values></valid_values>
     <default_value> </default_value>
     <values match="last" modifier="additive">
+      <!-- Turn on chemistry if not using physics testbed -->
+      <value compset="^((?!_CAM%PHYSTEST).)*$"> </value>
```
peverwhee, line 311: I think you can remove these lines.

peverwhee (changes requested):
Thanks @mattldawson !

small change and submodule question

ccpp_framework
nusbaume: That might have been me.  I thought we were supposed to commit the submodules (after running `git-fleximod update`) as well as the modification to `.gitmodules`, as otherwise commands like `git status` will continue showing that the repo is in an unclean state.

peverwhee: ahhh ok. yep. nevermind!

cime_config/config_component.xml
mattldawson: removed!

PR #288, opened 2024-08-14
Upgrade MPAS to version 8.2.1

kuanchihwang:
Currently, CAM-SIMA uses MPAS at commit `ed6f8e39ec0a811b6d079ca0fc6f9fb6e30bad23`, which is half way between MPAS versions 8.0.1 and 8.1.0.

This PR upgrades MPAS to version 8.2.1, which was released on August 7, 2024.

* MPAS dynamical core and its subdriver in CAM-SIMA now support MPI Fortran 2008 interface through the `mpi_f08` module. It can be enabled by defining the `MPAS_USE_MPI_F08` macro in `CPPFLAGS`.
* The log file (e.g., `log.atmosphere.0000.out`) produced by MPAS dynamical core now correctly indicates its git version string. Before, it would always show "N/A" in the log, which is not very helpful. Now, if the git submodule of MPAS is aligned at a tag, it will just show the tag name (e.g., "v8.2.1"). Otherwise, if the git submodule of MPAS is not aligned at a tag, it will show the nearest tag name, the number of commits after that tag and the abbreviated commit hash (e.g., "v8.0.1-54-ged6f8e39e").
* Tidy up the build directory of MPAS dynamical core.

### DIFF
--- a/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
+++ b/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
@@ -10,7 +10,12 @@ module dyn_mpas_subdriver
     use, intrinsic :: iso_fortran_env, only: output_unit
 
     ! Modules from external libraries.
+#ifdef MPAS_USE_MPI_F08
+    use mpi_f08, only: mpi_comm_null, mpi_comm_rank, mpi_success, &
+                       mpi_comm_type => mpi_comm, operator(==), operator(/=)
+#else
     use mpi, only: mpi_comm_null, mpi_comm_rank, mpi_success
+#endif
     use pio, only: pio_char, pio_int, pio_real, pio_double, &
                    file_desc_t, iosystem_desc_t, pio_file_is_open, pio_iosystem_is_active, &
                    pio_inq_varid, pio_inq_varndims, pio_inq_vartype, pio_noerr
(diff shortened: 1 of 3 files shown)


nusbaume (approved):
Looks good to me, thanks @kuanchihwang!

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -10,7 +10,12 @@ module dyn_mpas_subdriver
     use, intrinsic :: iso_fortran_env, only: output_unit
 
     ! Modules from external libraries.
+#ifdef MPAS_USE_MPI_F08
+    use mpi_f08, only: mpi_comm_null, mpi_comm_rank, mpi_success, &
+                       mpi_comm_type => mpi_comm, operator(==), operator(/=)
```
mgduda, line 15: Is the `/=` operator needed?

mgduda (changes requested):
Perhaps no real changes are needed, but just a question about the need for the `/=` operator from the `mpi_f08` module.

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
kuanchihwang: No, it is not needed. I have removed it in https://github.com/ESCOMP/CAM-SIMA/pull/288/commits/7e059d1f16b498035095a519f1dd244c19032049.

mgduda (approved):

Issue #123, opened 2021-04-27
CAMDEN should initialize tendencies


gold2718:
CAMDEN should initialize all tendencies in `physics_tend` to zero at the beginning of each physics timestep.
It would be nice if generate_registry could generate a routine to do this.

PR #289, opened 2024-08-19
Fix isssue-closing script to work without project pages and default merge messages.

### LINKED ISSUE #279: Issue closing script doesn't work with PR description commit
### What happened?

The CAM-SIMA repo was recently modified to commit the entire PR title and description with the merge commit, instead of just the text:

`Merge pull request #XXX`

However, the issue-closing Github Action keys off of the above text.  Thus it currently won't work with the new PR commit message provided by Github.  

To fix this issue, the script should be modified to key off different text, ideally provided by a PR template so that users don't need to add it to the PR manually.

 

Merge a PR into the `development` branch with one of the issue-closing keywords in the PR description.  One will find that the issue does not actually close.

4609b76

Other (please explain below)

Other (please specify below)

_No response_

### Extra info

_No response_

nusbaume:
Fixes the issue-closing script in CAM-SIMA so that it will now successfully close issues or PRs in the non-default branch when issues are referenced in the PR message with a relevant keyword.  This will happen both without any attachment to a particular Github project, and with any PR commit message.

Fixes #279 

List any changes to the defaults for the boundary datasets: N/A

Describe any substantial timing or memory changes: N/A

Code reviewed by: peverwhee, mwaxmonsky 

List all existing files that have been modified, old files eliminated, new files added and describe the changes:

M .github/workflows/branch_push_workflow.yml
    - Update Github Action and Python versions.

M .github/scripts/branch_pr_issue_closer.py
    - Update issue-closing script to work without projects or default PR commit messages.

derecho/intel/aux_cam: All pass.
derecho/gnu/aux_cam:  All pass.

Summarize any changes to answers, i.e.,

what code configurations: N/A
what platforms/compilers: N/A
nature of change (roundoff; larger than roundoff but same climate; new climate): N/A

### DIFF
--- a/.github/scripts/branch_pr_issue_closer.py
+++ b/.github/scripts/branch_pr_issue_closer.py
@@ -228,7 +228,7 @@ def _main_prog():
         #Create new "closed" PR list:
         close_pulls = []
 
-        #Search text right after keywords for possible issue numbers:
+        #Go through all matches to pull out PR and issue numbers:
         for match in word_matches:
 
             issue_dict = match.groupdict()
--- a/.github/scripts/branch_pr_issue_closer.py
+++ b/.github/scripts/branch_pr_issue_closer.py
@@ -26,6 +26,25 @@ import argparse
 from github import Github
 from github import Auth
 
+###############
+#REGEX PATTERNS
+###############
+
+#Issue-closing Keywords are:
+#close, closes, closed
+#fix, fixes, fixed
+#resolve, resolves, resolved
+
+#Create relevant regex patterns:
+_CLOSE_KEY              = r'close[sd]?'
+_FIX_KEY                = r'fix(e[sd])?'
+_RESOLVE_KEY            = r'resolve[sd]?'
+_KEYWORDS               = rf'{_CLOSE_KEY}|{_FIX_KEY}|{_RESOLVE_KEY}'
+_KEYWORDS_CAPTURE_GROUP = rf'(?P<keyword>{_KEYWORDS})'
+_ID_NUMBER              = r'\d+'
+_ID_CAPTURE_GROUP       = rf'(?P<id>{_ID_NUMBER})'
+_LINKED_ISSUE_PATTERN   = rf'{_KEYWORDS_CAPTURE_GROUP}\s*#{_ID_CAPTURE_GROUP}'
+
 #################
 #HELPER FUNCTIONS
 #################
@@ -184,35 +203,24 @@ def _main_prog():
         #Check if one of the keywords exists in PR message
         #+++++++++++++++++++++++++++++++++++++++++++++++++
 
-        #Keywords are:
-        #close, closes, closed
-        #fix, fixes, fixed
-        #resolve, resolves, resolved
-
-        #Create regex pattern to find keywords:
-        keyword_pattern = re.compile(r'(^|\s)close(\s|s\s|d\s)|(^|\s)fix(\s|es\s|ed\s)|(^|\s)resolve(\s|s\s|d\s)')
+        #Compile regex patterns into object:
+        keyword_pattern = re.compile(_LINKED_ISSUE_PATTERN)
 
         #Extract (lower case) Pull Request message:
         pr_msg_lower = merged_pull.body.lower()
 
-        #search for at least one keyword:
-        word_matches = []
-        if keyword_pattern.search(pr_msg_lower) is not None:
-            #If at least one keyword is found, then determine location of every keyword instance:
-            word_matches = keyword_pattern.finditer(pr_msg_lower)
-        else:
+        #End script if no keywords found:
+        if keyword_pattern.search(pr_msg_lower) is None:
             endmsg = f"Pull request #{pr_num} was merged without using any of the keywords. "
             endmsg += "Thus there are no issues to close."
             end_script(endmsg)
 
+        #search for at least one keyword in PR message:
+        word_matches = keyword_pattern.finditer(pr_msg_lower, re.IGNORECASE)
+
         #++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
         #Extract issue and PR numbers associated with found keywords in merged PR message
         #++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-
-        #create issue pattern ("the number symbol {#} + a number"),
-        #which ends with either a space, a comma, a period, or
-        #the end of the string itself:
-        issue_pattern = re.compile(r'#[0-9]+(\s|,|$)|.')
 
         #Create new "close" issues list:
         close_issues = []
@@ -223,38 +231,16 @@ def _main_prog():
         #Search text right after keywords for possible issue numbers:
         for match in word_matches:
 
-            #create temporary string starting at end of match:
-            tmp_msg_str = pr_msg_lower[match.end():]
+            issue_dict = match.groupdict()
+            issue_num  = int(issue_dict['id'].lstrip('0'))
 
-            #Check if first word matches issue pattern:
-            if issue_pattern.match(tmp_msg_str) is not None:
-
-                #If so, then look for an issue number immediately following
-                first_word = tmp_msg_str.split()[0]
-
-                #Extract issue number from first word:
-                try:
-                    #First try assuming the string is just a number
-                    issue_num = int(first_word[1:]) #ignore "#" symbol
-                except ValueError:
-                    #If not, then ignore the last character, in case the user
-                    #included punctutation (i.e. a space, comma, or period)
-                    #after the PR number:
-                    try:
-                        issue_num = int(first_word[1:-1])
-                    except ValueError:
-                        #If ignoring the first and last letter doesn't work,
-                        #then the match was likely a false positive,
-                        #so set the issue number to one that will never be found:
-                        issue_num = -9999
-
-                #Check if number is actually for a PR (as opposed to an issue):
-                if issue_num in open_pulls:
-                    #Add PR number to "close pulls" list:
-                    close_pulls.append(issue_num)
-                elif issue_num in open_issues:
-                    #If in fact an issue, then add to "close issues" list:
-                    close_issues.append(issue_num)
+            #Check if number is actually for a PR (as opposed to an issue):
+            if issue_num in open_pulls:
+                #Add PR number to "close pulls" list:
+                close_pulls.append(issue_num)
+            elif issue_num in open_issues:
+                #If in fact an issue, then add to "close issues" list:
+                close_issues.append(issue_num)
 
     #+++END REFERENCED PR LOOP+++
 
--- a/.github/scripts/branch_pr_issue_closer.py
+++ b/.github/scripts/branch_pr_issue_closer.py
@@ -35,6 +35,12 @@ from github import Auth
 #fix, fixes, fixed
 #resolve, resolves, resolved
 
+#The keywords are designed to match
+#the keywords that exist in Github
+#already for default branches, which
+#can be found here:
+#https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue
+
 #Create relevant regex patterns:
 _CLOSE_KEY              = r'close[sd]?'
 _FIX_KEY                = r'fix(e[sd])?'
--- a/.github/scripts/branch_pr_issue_closer.py
+++ b/.github/scripts/branch_pr_issue_closer.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
 """
 Script name:  branch_PR_issue_closer.py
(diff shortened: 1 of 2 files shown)


.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,162 @@ def _main_prog():
...
-    else:
-        endmsg = "Pull request was merged without using any of the keywords.  Thus there are no issues to close."
-        end_script(endmsg)
+        #search for at least one keyword:
+        word_matches = []
+        if keyword_pattern.search(pr_msg_lower) is not None:
+            #If at least one keyword is found, then determine location of every keyword instance:
+            word_matches = keyword_pattern.finditer(pr_msg_lower)
+        else:
+            endmsg = "Pull request was merged without using any of the keywords.  Thus there are no issues to close."
```
peverwhee, line 204: Might be nice to include the PR number
```suggestion
            endmsg = f"Pull request #{pr_num} was merged without using any of the keywords.  Thus there are no issues to close."
```

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,162 @@ def _main_prog():
...
 
-        try:
-            #Try assuming the word is just a number:
-            pr_num = int(first_word[1:]) #ignore "#" symbol
-        except ValueError:
-            #If the conversion fails, then this is likely not a real PR merge, so end the script:
-            endmsg = "No Pull Request number was found in the commit message, so there is nothing for the script to do."
+        #If pull request has not been merged, then exit script:
+        if not merged_pull.merged:
+            endmsg = f"Pull request associated with commit:\n{trigger_sha}\n"
```
peverwhee, line 143: Might be nice to include PR number
```suggestion
            endmsg = f"Pull request #{pr_num} associated with commit:\n{trigger_sha}\n"
```

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,162 @@ def _main_prog():
...
+        #Extract merged branch from latest Pull request:
+        merged_branch = merged_pull.base.ref
 
-    #If PR was to default branch, then exit script (as github will handle it automatically):
-    if merged_branch == default_branch:
-        endmsg = "Pull request ws merged into default repo branch. Thus issue is closed automatically"
-        end_script(endmsg)
+        #If PR was to default branch, then exit script (as github will handle it automatically):
+        if merged_branch == default_branch:
+            endmsg = "Pull request was merged into default repo branch. "
```
peverwhee, line 159: Might be nice to include the PR number
```suggestion
            endmsg = f"Pull request #{pr_num} was merged into default repo branch. "
```

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,162 @@ def _main_prog():
...
-            #Check if number is actually for a PR (as opposed to an issue):
-            if issue_num in open_pulls:
-                #Add PR number to "close pulls" list:
-                close_pulls.append(issue_num)
-            elif issue_num in open_issues:
-                #If in fact an issue, then add to "close issues" list:
-                close_issues.append(issue_num)
+                    #If not, then ignore last letter:
+                    try:
+                        issue_num = int(first_word[1:-1])
```
peverwhee, line 241: I know this isn't new with this PR, but what is the use case for this? (why ignore the last letter?)

peverwhee (commented):
A question and a couple of suggestions.

.github/scripts/branch_pr_issue_closer.py
nusbaume: Good call!  PR number added.

.github/scripts/branch_pr_issue_closer.py
nusbaume: Yes!  PR number added.

.github/scripts/branch_pr_issue_closer.py
nusbaume: Yes!  PR number added.

.github/scripts/branch_pr_issue_closer.py
nusbaume: This is because the regex used to extract the `first_word` variable allows for it to end with a space, period, or comma, and so for situations where there is a comma or period you'll want to ignore the last character.  I believe this was done so that users could write lists like the following in their PR description:

`Fixes #X, Closes #Y, Resolves #Z.`

Of course if you think instead we should require only a space (or nothing) at the end then I can change the regex and remove this `try` statement.  Just let me know!

peverwhee: ahhh punctuation! of course!

that makes sense - thanks. Can you perhaps add a comment explaining that for the next person (possibly me again) who is confused?

nusbaume: Ahh yes, I have heard that those "comment" things you mentioned are useful.  Comment added!

peverwhee: thank you!!

peverwhee (approved):

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,165 @@ def _main_prog():
...
-    #resolve, resolves, resolved
+        #Keywords are:
+        #close, closes, closed
+        #fix, fixes, fixed
+        #resolve, resolves, resolved
 
-    #Create regex pattern to find keywords:
-    keyword_pattern = re.compile(r'(^|\s)close(\s|s\s|d\s)|(^|\s)fix(\s|es\s|ed\s)|(^|\s)resolve(\s|s\s|d\s)')
+        #Create regex pattern to find keywords:
+        keyword_pattern = re.compile(r'(^|\s)close(\s|s\s|d\s)|(^|\s)fix(\s|es\s|ed\s)|(^|\s)resolve(\s|s\s|d\s)')
```
mwaxmonsky, line 193:
```suggestion
        _CLOSE_KEY      = r'close(s|d)?'
        _FIX_KEY        = r'fix(e(s|d))?'
        _RESOLVE_KEY    = r'resolve(s|d)?'
        _KEYWORDS       = rf'{_CLOSE_KEY}|{_FIX_KEY}|{_RESOLVE_KEY}'
        keyword_pattern = re.compile(r'(^|\s)({_KEYWORDS})\s')
```
I know this was already here but it might be worth while to break up the keywords to make it easier to track the whitespace checks as well.

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,165 @@ def _main_prog():
...
-        word_matches = keyword_pattern.finditer(pr_msg_lower)
-    else:
-        endmsg = "Pull request was merged without using any of the keywords.  Thus there are no issues to close."
-        end_script(endmsg)
+        #search for at least one keyword:
+        word_matches = []
+        if keyword_pattern.search(pr_msg_lower) is not None:
+            #If at least one keyword is found, then determine location of every keyword instance:
+            word_matches = keyword_pattern.finditer(pr_msg_lower)
+        else:
```
mwaxmonsky, line 203:
```suggestion
        word_matches = keyword_pattern.finditer(pr_msg_lower)
        if word_matches is None:
```
Is there any benefit to only doing a `search` first if we have to `finditer` anyway if there's a match?  I imagine this will scan the entire message string anyway as GitHub defaults to putting the keywords at the bottom of the PR message.

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,165 @@ def _main_prog():
...
+        #++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-    #create issue pattern ("the number symbol {#} + a number"),
-    #which ends with either a space, a comma, a period, or
-    #the end of the string itself:
-    issue_pattern = re.compile(r'#[0-9]+(\s|,|$)|.')
+        #create issue pattern ("the number symbol {#} + a number"),
+        #which ends with either a space, a comma, a period, or
+        #the end of the string itself:
+        issue_pattern = re.compile(r'#[0-9]+(\s|,|$)|.')
```
mwaxmonsky, line 215: Happy to add this myself (and again, I completely understand this isn't the goal with this PR), but can we add this to the above PR keyword as a group?  That way we get both the keyword and number in one search without needing to have extra whitespace/special character checking.

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,165 @@ def _main_prog():
...
-            try:
-                #First try assuming the string is just a number
-                issue_num = int(first_word[1:]) #ignore "#" symbol
-            except ValueError:
-                #If not, then ignore last letter:
+                #Extract issue number from first word:
                 try:
-                    issue_num = int(first_word[1:-1])
+                    #First try assuming the string is just a number
+                    issue_num = int(first_word[1:]) #ignore "#" symbol
```
mwaxmonsky, line 238: Ah, I lied, the group change would simplify this as well.

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,176 +98,165 @@ def _main_prog():
...
+                    #If not, then ignore the last character, in case the user
+                    #included punctutation (i.e. a space, comma, or period)
+                    #after the PR number:
+                    try:
+                        issue_num = int(first_word[1:-1])
+                    except ValueError:
+                        #If ignoring the first and last letter doesn't work,
+                        #then the match was likely a false positive,
+                        #so set the issue number to one that will never be found:
+                        issue_num = -9999
```
mwaxmonsky, line 249: ... And would mitigate the need for this as well.

mwaxmonsky (changes requested):
One or two things I think might be worth while and one change I'm happy to help look into a bit more to make the number tracking simpler.

Great work @nusbaume!

.github/scripts/branch_pr_issue_closer.py
mwaxmonsky:
```python
import re

_CLOSE_KEY   = r'close[sd]?'
_FIX_KEY     = r'fix(e[sd])?'
_RESOLVE_KEY = r'resolve[sd]?'
_KEYWORDS = rf'{_CLOSE_KEY}|{_FIX_KEY}|{_RESOLVE_KEY}'
_ID_NUMBER = r'[1-9]\d*'
keyword_pattern = rf'(?P<keyword>{_KEYWORDS})\s*#(?P<id>{_ID_NUMBER})'
to_search = 'closed #2, Fixes #4'

for m in re.finditer(keyword_pattern, to_search, re.IGNORECASE):
    issue_dict = m.groupdict()
    print(issue_dict['keyword'], issue_dict['id'])
# prints:
# closed 2
# Fixes 4
```
@nusbaume, this should make the search code a little easier and also works for multiline strings.

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -21,52 +21,34 @@
...
+
+#Issue-closing Keywords are:
+#close, closes, closed
+#fix, fixes, fixed
+#resolve, resolves, resolved
+
+#Create relevant regex patterns:
+_CLOSE_KEY              = r'close[sd]?'
+_FIX_KEY                = r'fix(e[sd])?'
+_RESOLVE_KEY            = r'resolve[sd]?'
```
mwaxmonsky, line 41: Might be worth mentioning the source webpage these get pulled from in case they change or more are added later:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

.github/scripts/branch_pr_issue_closer.py
```diff
@@ -135,168 +117,122 @@ def _main_prog():
...
+        #Extract (lower case) Pull Request message:
+        pr_msg_lower = merged_pull.body.lower()
 
-    #Create new "close" issues list:
-    close_issues = []
+        #End script if no keywords found:
+        if keyword_pattern.search(pr_msg_lower) is None:
+            endmsg = f"Pull request #{pr_num} was merged without using any of the keywords. "
+            endmsg += "Thus there are no issues to close."
+            end_script(endmsg)
```
mwaxmonsky, line 216: Since we are now using the `finditer` function, would it make sense to move this after the for loop and use `if len(close_issues) == 0 and len(close_pulls) == 0)` to avoid scanning the description twice?

Don't suspect this is a hot section of code since it runs on GitHub but if someone writes a book of a description, it might matter?  But if these reads cleaner, please disregard!

mwaxmonsky (approved):
Looks great @nusbaume, thanks for taking this on!

One or two minor suggestions but nothing that should hold this up any longer.

.github/scripts/branch_pr_issue_closer.py
nusbaume: Good call!  Comment with URL has been added.

.github/scripts/branch_pr_issue_closer.py
nusbaume: While this would certainly make the code more performant, the current configuration allows me to have different return messages depending on which step "fails", which I have found handy as sometimes developers come to me telling me the script doesn't work when in fact it was just user error when writing the PR description.  

Of course if you think it would still be worth getting the performance boost then just let me know, as we can always make a new issue and tackle it at a later date.  Thanks!

mwaxmonsky: I'm not sure I follow.  The error message and the context would still be the same so that wouldn't change, it's just the location we'd be doing it at, no?

nusbaume: Not quite, because that `search` check is to see if you have the correct syntax in the PR description, e. g. `Fixes #123`.  However, after the iteration loop there is a check on the `close_x` lists to see if any of the found issue numbers are actually open, and if not it states that no open issue or PR numbers were found.

In other words it checks that even if your PR description syntax is correct, that the number you are using is valid.  See here:

https://github.com/ESCOMP/CAM-SIMA/blob/development/.github/scripts/branch_pr_issue_closer.py#L253

Hopefully that makes sense?  If not just let me know!

nusbaume: Whoops, wrong code quote, I meant start looking here:

https://github.com/ESCOMP/CAM-SIMA/blob/development/.github/scripts/branch_pr_issue_closer.py#L243

PR #290, opened 2024-08-21
PR template

peverwhee:
Brings in the new PR template for CAM-SIMA.

See: https://github.com/peverwhee/CAM-SIMA/pull/2 for what it looks like upon opening a PR (though it's not wildly different from the markdown...).


.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
+Tag name (required for main):
```
nusbaume, line 1: I would change this to say `release branches`:

```suggestion
Tag name (required for release branches):
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
...
+List any changes to the defaults for the input datasets (e.g. boundary datasets):
+
+List all files eliminated and why:
+
+List all files added and what they do:
+
+List all existing files that have been modified, and describe the changes: 
+(Helpful git command: `git diff --name-status development...<your_branch_name>`)
+
+If there are new failures (compare to the test/existing-test-failures.txt file),
```
nusbaume, line 19: I might vote to have the file listed as "code" here:

```suggestion
If there are new failures (compare to the `test/existing-test-failures.txt` file),
```

nusbaume (approved):
This PR template looks good to me!  Just a couple small requests.

.github/pull_request_template.md
peverwhee: changed!

.github/pull_request_template.md
peverwhee: changed!

.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
+Tag name (required for release branches):
```
cacraigucar, line 1: Maybe modify to say `(required for commits which will be moved immediately to release branch)`

.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
...
+Originator(s):
+
```
cacraigucar, line 3: Do we want a `"Description"` section which is what the user thinks might be the overall summary of the PR when they open it.  Maybe the title captures this, but I'm not sure.  I'm trying to think about the two stages for this description, what is known when the PR is opened and what it is at the very end, right before merge.

.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
...
+Summary (include the keyword ['closes', 'fixes', 'resolves'] and issue number):
```
cacraigucar, line 4: I believe we should include the issue title as well (a pain to copy/type, but very important for getting a summary of what issue xyz is all about)

.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
...
+List any changes to the defaults for the input datasets (e.g. boundary datasets):
+
+List all files eliminated and why:
+
+List all files added and what they do:
+
+List all existing files that have been modified, and describe the changes: 
+(Helpful git command: `git diff --name-status development...<your_branch_name>`)
+
+If there are new failures (compare to the `test/existing-test-failures.txt` file),
```
cacraigucar, line 19: Maybe change `compare `=> `compared` ?  I can see it either way

cacraigucar (changes requested):

.github/pull_request_template.md
peverwhee: @cacraigucar @nusbaume how do we feel about changing "Summary" to "Description"? I think maybe "Summary" is what the PR title is and "Description" opens the door for more detail?

.github/pull_request_template.md
peverwhee: I *think* the `(required for release branches)` is for PRs to main (or a release branch), rather then for PRs that are targeted for development but are destined for release. Does that make sense?

.github/pull_request_template.md
peverwhee: done!

.github/pull_request_template.md
peverwhee: done! i agree that "compared" is a bit better

.github/pull_request_template.md
nusbaume: I am happy with changing `Summary` to `Description` in the template.

cacraigucar: Sounds good to me

peverwhee: change made!

.github/pull_request_template.md
cacraigucar: I think that the commits happen on the development branch and then the main branch doesn't actually have commits, but rather merges.  Perhaps I'm thinking of a different repo's workflow where this happens.  @nusbaume - where did Haipeng have his commit not having the tagname because it was committed in the branch and he didn't have an opportunity to insert it later.

jimmielin: It was in atmospheric physics I think so the tag name had to be filled in by a future commit.

nusbaume: Yeah I ended up filling in @jimmielin's tag name later.  I agree with @cacraigucar's wording, maybe we can change it slightly to something like `(required for the last development PR before merging into a release branch)`?

nusbaume: Actually scratch that last comment.  I think we can leave it as just `(required for release branches)` because for CAM-SIMA the release branch will always have a PR template entry, even for merge commits, so we can just add the tag name in then.

cacraigucar (approved):

nusbaume:
Merging this now, with the assumption that it can always be modified later if folks have any future suggestions or comments.

PR #295, opened 2024-08-26
Fix user_defined physical constants not updated outside of masterproc

jimmielin:
`user_defined` physical constants in `physconst.F90` are not saved outside of `masterproc` since the in-module variable assignments were made inside a `if(masterproc)`


jimmielin:
All working after a clean rebuild, not sure why `fleximod-test` is still stuck though...

nusbaume (approved):

cacraigucar (approved):

PR #310, opened 2024-10-08
Fix CCPP Fortran parser / build error with tropopause_find

jimmielin:
The CCPP Fortran parser is not able to read module variables of character type defined with the `shr_kind_cl` parameter size.

The `tropopause_climo_file` variable does not need to be a CCPP variable anyway so it has been removed. This will fix the build process again.

Verified that it builds (need to either remove `tropopause_diagnostics` from tropopause_find test SDF -or- merge the history PR) and runs with CAM snapshot.

### DIFF
--- a/src/physics/utils/tropopause_climo_read.F90
+++ b/src/physics/utils/tropopause_climo_read.F90
@@ -17,6 +17,9 @@ module tropopause_climo_read
   public :: tropopause_climo_readnl
   public :: tropopause_climo_read_file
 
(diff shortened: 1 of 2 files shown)


jimmielin:
OK I have confirmed this time that with all the modifications all tests pass on Derecho (except the submit step)!

nusbaume (approved):

peverwhee (approved):

PR #313, opened 2024-10-16
New const_get_index logic without cam_ccpp_cap dependency

jimmielin:
This avoids a circular dependency if schemes require `const_get_index`, which presently depends on `cam_ccpp_cap`, which depends on individual schemes.


jimmielin:
Closing as alternative in `to_be_ccppized` was discussed and will be implemented.

PR #314, opened 2024-10-16
New const_get_index logic without cam_ccpp_cap dependency (in atmos_phys)

jimmielin:
Companion atmospheric_physics PR: https://github.com/ESCOMP/atmospheric_physics/pull/135

This avoids a circular dependency if schemes require const_get_index, which presently depends on cam_ccpp_cap, which depends on individual schemes.

### DIFF
--- a/cime_config/cam_autogen.py
+++ b/cime_config/cam_autogen.py
@@ -602,7 +602,7 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
         if not os.path.isdir(atm_phys_util_dir):
             #CAM-SIMA will likely not run without this, so raise an error
             emsg = "ERROR: Unable to find CCPP physics utilities directory:\n"
-            emsg += f" {atm_phys_util_dir}\n Have you run 'checkout_externals'?"
+            emsg += f" {atm_phys_util_dir}\n Have you run 'git-fleximod'?"
             raise CamAutoGenError(emsg)
         # end if
 
@@ -611,6 +611,16 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
         for util_file in atm_phys_util_files:
             shutil.copy(util_file, physics_blddir)
         # end for
+
+        # Copy to_be_ccppized utility modules to the build directory,
+        # as SIMA cam_constituents depends on them.
+        atm_phys_to_be_ccppized_dir = os.path.join(atm_phys_top_dir, "to_be_ccppized")
+
+        # Check that the directory exists and copy to build directory
+        if os.path.isdir(atm_phys_to_be_ccppized_dir):
+             atm_phys_to_be_ccppized_files = glob.glob(os.path.join(atm_phys_to_be_ccppized_dir, "*.F90"))
+             for to_be_ccppized_file in atm_phys_to_be_ccppized_files:
+                shutil.copy(to_be_ccppized_file, physics_blddir)
     # end if
 
     if do_gen_ccpp or do_gen_nl:
--- a/cime_config/cam_autogen.py
+++ b/cime_config/cam_autogen.py
@@ -600,7 +600,7 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
 
         # Check that directory exists
         if not os.path.isdir(atm_phys_util_dir):
-            #CAM-SIMA will likely not run without this, so raise an error
+            # CAM-SIMA will likely not run without this, so raise an error
             emsg = "ERROR: Unable to find CCPP physics utilities directory:\n"
             emsg += f" {atm_phys_util_dir}\n Have you run 'git-fleximod'?"
             raise CamAutoGenError(emsg)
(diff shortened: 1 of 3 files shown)


cime_config/cam_autogen.py
```diff
@@ -611,6 +611,16 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
...
+
+        # Copy to_be_ccppized utility modules to the build directory,
+        # as SIMA cam_constituents depends on them.
+        atm_phys_to_be_ccppized_dir = os.path.join(atm_phys_top_dir, "to_be_ccppized")
+
+        # Check that the directory exists and copy to build directory
+        if os.path.isdir(atm_phys_to_be_ccppized_dir):
+             atm_phys_to_be_ccppized_files = glob.glob(os.path.join(atm_phys_to_be_ccppized_dir, "*.F90"))
+             for to_be_ccppized_file in atm_phys_to_be_ccppized_files:
+                shutil.copy(to_be_ccppized_file, physics_blddir)
```
cacraigucar, line 623: Here is what someone gave me for this code section (an alternate implementation?)  Note that the code capture isn't capturing the whole section for some reason.

`#WARNING:  THIS SHOULD BE DELETED ONCE DEPENDENCIES ARE ENABLED!!!!!! - CHERYL AND JESSE
    ########################################################################################
    if do_gen_ccpp:
        # Set CCPP physics "to_be_ccppized" path
        atm_phys_tobe_dir = os.path.join(atm_phys_src_dir, "to_be_ccppized")

        # Check that directory exists
        if not os.path.isdir(atm_phys_tobe_dir):
            #CAM-SIMA will likely not run without this, so raise an error
            emsg = "ERROR: Unable to find CCPP physics 'to_be_ccppized' directory:\n"
            emsg += f" {atm_phys_tobe_dir}\n Have you run 'checkout_externals'?"
            raise CamAutoGenError(emsg)
        # end if

        # Copy all "to_be_ccppized" source files to the build directory
        atm_phys_tobe_files = glob.glob(os.path.join(atm_phys_tobe_dir, "*.F90"))
        for tobe_file in atm_phys_tobe_files:
            shutil.copy(tobe_file, physics_blddir)
        # end for
    # end if
    ###########################################################################################

`

jimmielin: Thanks @cacraigucar ! It looks like the same logic except for the error if to_be_ccppized directory is not present - I suppose if it passed the check for `schemes/utilities` to be present then `git-fleximod` has already been ran, so I just checked if the directory exists and copy its contents.

I'm not sure of "once dependencies are enabled" mean - would the dependencies code help SIMA discover `to_be_ccppized/ccpp_const_utils.F90` if a CAM module depends on it? I suppose I could add the dependency to the `cam_constituents.meta` file but I don't know if it's wired to the SIMA build system.

nusbaume: I think this was code I gave to you @cacraigucar when we needed to get around the lack of the ability to use the CCPP dependencies method in CAM-SIMA.  However, given that now that particular issue been fixed we should probably change the warning to instead just say something about being removed once the CCPP constituents object is directly accessible.

Also, I think the `isdir` path check is good to have since it is only done once and it will probably help us quickly debug issues if we ever reach the situation where we move, delete, or rename that directory in atmospheric_physics.

Anyways, that's my two cents on the topic, but am happy to hear alternative arguments!

jimmielin: Thanks @nusbaume! I'll change this.

I guess the other question I had is that if it's OK to keep `ccpp_const_utils.F90` in `to_be_ccppized/` and copy all `to_be_ccppized/*.F90` to be in `bld`. To avoid breaking future builds if other files get put in, I could move it to `to_be_ccppized/constituents` for safety - please let me know what you think!

nusbaume: Good point!  I think your idea of just copying over a `to_be_ccppized/constituents` subdirectory would be the best solution, at least for our current set of use cases.

cacraigucar: I'm curious why you think we need to have subfolders in to_be_ccppized.  I think the hope is that we will not have this folder "long-term", but rather is a catch-all folder to hold the items we need to convert to CCPP when we have time.  I'd think a flat structure would be fine and would give us a quick look at what we still need to work on in this directory.

jimmielin: Hi @cacraigucar, I think I wasn't sure if I wanted to copy everything in `to_be_ccppized` to the build directory regardless of the physics suite. Michael's vertical diffusion solver PR also included `utilities` and `utilities/vertical_diffusion` so I took inspiration from that and decided to add subdirectories. But I'm open to any options!

Do you use the code you shared above for ZM to handle adding to_be_ccppized files to the build list? Will it be a problem to add `to_be_ccppized/*.F90` to the build list at all times? Thanks!

jimmielin: Hi @cacraigucar, @nusbaume - after thinking about this a bit more I agree with Cheryl to keep `to_be_ccppized` as flat as possible for the reasons mentioned, and also because I think having `to_be_ccppized/constituents` and a whole set of other folders that probably contain one (short-term) file doesn't make it more organized than enumerating all the source files in `cam_autogen` if need be, since we are hoping these are just temporary solutions.

For this reason I've kept `ccpp_const_utils.F90` in just `to_be_ccppized` for now, but I updated the logic to report error when the folder is not present, and updated the comments that this functionality and future ones has to be moved to SIMA framework or properly CCPPized in the future. Please let me know if that works. Thanks!

cacraigucar (approved):

cime_config/cam_autogen.py
```diff
@@ -611,6 +611,25 @@ def generate_physics_suites(build_cache, preproc_defs, host_name,
...
+        # as SIMA cam_constituents depends on them.
+        # Note: to_be_ccppized utility modules to be removed once functionality is migrated
+        # to SIMA or CCPPized in atmospheric_physics.
+        atm_phys_to_be_ccppized_dir = os.path.join(atm_phys_top_dir, "to_be_ccppized")
+
+        # Check that the directory exists
+        if not os.path.isdir(atm_phys_to_be_ccppized_dir):
+            # CAM-SIMA will likely not run without this, so raise an error
+            emsg = "ERROR: Unable to find CCPP physics to_be_ccppized directory:\n"
+            emsg += f" {atm_phys_util_dir}\n Have you run 'git-fleximod'?"
```
nusbaume, line 625: I think this should be `atm_phys_to_be_ccppized_dir` here:

```suggestion
            emsg += f" {atm_phys_to_be_ccppized_dir}\n Have you run 'git-fleximod'?"
```

src/physics/utils/cam_constituents.F90
```diff
@@ -312,10 +316,14 @@ subroutine const_get_index(name, cindex, abort, warning, caller)
       character(len=*), parameter :: subname = 'const_get_index: '
       !-----------------------------------------------------------------------
 
-      ! Find tracer name in the master table
-      call cam_const_get_index(name, cindex, errcode=errcode, errmsg=errmsg)
+      call ccpp_const_get_idx(const_props, name, cindex, errmsg, errcode)
 
       if (errcode /= 0) then
+         call endrun(subname//"Error "//to_str(errcode)//": "//           &
```
nusbaume, line 322: I might use `stringify` here instead of `to_str` (my hope is to eventually use `stringify` for all string conversions):

```suggestion
         use string_utils, only: stringify
         ...
         call endrun(subname//"Error "//stringify((/errcode/))//": "//           &
```

nusbaume (changes requested):
Looks good, but did have two small change requests.

cime_config/cam_autogen.py
jimmielin: Oops, my bad! Fixed

src/physics/utils/cam_constituents.F90
jimmielin: Thanks, updated!

jimmielin:
Thanks @nusbaume! Apologies for the oversight. Updated and waiting for tests to run then I'll merge to development.

jimmielin:
All build tests passed - had to wait in the queue for longer than I expected. Merging now! Thanks everyone for the reviews and feedback.

PR #342, opened 2024-12-18
Fix water vapor (Q) register incompatibility with initialize_constituents scheme

jimmielin:
Originator(s): @jimmielin

Add `water_species` and `mixing_ratio_type = 'wet'` property for when `cam_register_constituents` registers water vapor when no other scheme explicitly uses water vapor (but might use all CCPP constituents as a whole)

```
Add water_species = .true. and mixing_ratio_type = 'wet' to wv_stdname
M       src/control/cam_comp.F90
```

derecho/intel/aux_sima:

derecho/gnu/aux_sima:

### DIFF
--- a/src/control/cam_comp.F90
+++ b/src/control/cam_comp.F90
@@ -615,6 +615,8 @@ CONTAINS
               default_value=0._kind_phys,                                               &
               vertical_dim="vertical_layer_dimension",                                  &
               advected=.true.,                                                          &
+              water_species = .true.,                                                   &
+              mixing_ratio_type = 'wet',                                                &
            errcode=errflg, errmsg=errmsg)
 
          if (errflg /= 0) then


jimmielin:
When testing shallow convection with the `initialize_constituents` scheme to use all constituent data, I have found that the model fails to initialize with the error
```
 ERROR:
 cam_register_constituents: ccp_model_const_add_metadata ERROR: Trying to add co
 nstituent water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water but an inc
 ompatible constituent with this name already exists at cam_comp.F90:638
```

Debugging in the CCPP constituents property module shows that the following logic in `cam_register_constituents`:
```fortran
      ! Check if water vapor is already marked as a constituent by the
      ! physics:
      call cam_ccpp_is_scheme_constituent(wv_stdname, is_constituent, errflg, errmsg)
...

      !If not requested by the physics, then add water vapor to the
      !constituents object:
      !-------------------------------------------
      if (.not. is_constituent) then
...

         ! Register the constituents so they can be advected:
         call host_constituents(1)%instantiate( &
              std_name=wv_stdname,              &
              long_name="water vapor mixing ratio w.r.t moist air and condensed_water", &
...
```

Does not add the `water_species` property whereas `initialize_constituents` does for `Q`, leading to the runtime crash. As water vapor is a water species, these properties are now added which allows the model to proceed.

peverwhee (approved):
yes - good call @jimmielin !

nusbaume (approved):
Thanks for fixing this @jimmielin!

jimmielin:
Superseded by https://github.com/ESCOMP/atmospheric_physics/pull/185

PR #345, opened 2025-01-03
Updates for ZM round 4

cacraigucar:
Tag name (required for release branches): sima0_02_000zh
Originator(s): cacraig, peverley, nusbaume

- No issue made
- Adds mods from various developers to get ZM to work properly in CAM-SIMA

A       cime_config/testdefs/testmods_dirs/cam/outfrq_zm_derecho/shell_commands
A       cime_config/testdefs/testmods_dirs/cam/outfrq_zm_derecho/user_nl_cam
             - setup ZM regression test
             
A       src/physics/utils/ppgrid.F90
            - temporary file to use for backwards compatibility with CAM

M       .gitmodules
             - update to get atmospheric_physics changes for ZM
            
M       cime_config/atm_in_paramgen.py
             - exit if namelist entry does not have a value
             
M       cime_config/testdefs/testlist_cam.xml
             - add test for ZM
             
M       src/data/physconst.meta
M       src/data/ref_pres.meta
             - updated CCPP standard names
             
M       src/data/registry.xml
             - add entries required for ZM
            
M       src/physics/ncar_ccpp
             - updated submodule
             
M       src/physics/utils/musica_ccpp_dependencies.F90
M       src/physics/utils/musica_ccpp_dependencies.meta
             - remove cloud_area_fraction 
            
M       src/physics/utils/phys_comp.F90
             - Add new dimension: number_of_ccpp_constituents
             
M       src/utils/time_manager.F90
             - add current_timestep_number as "mark_as_initialized"
            
M       test/unit/python/test_atm_in_paramgen.py
             - @nusbaume added unit test for new python code

derecho/intel/aux_sima:
    - Added ZM test

derecho/gnu/aux_sima:
    - All pass


src/data/registry.xml
```diff
@@ -491,5 +491,76 @@
...
+              units="s" type="real" kind="kind_phys">
+    </variable>
+    <variable local_name="tpert"
+              standard_name="convective_temperature_perturbation_due_to_pbl_eddies"
+              units="K" type="real" kind="kind_phys"
+              allocatable="allocatable">
+      <dimensions>horizontal_dimension</dimensions>
+      <ic_file_input_names>tpert pbuf_tpert</ic_file_input_names>
+    </variable>
+    <variable local_name="ql"
```
jimmielin, line 524: Could I ask why the local name in SIMA is set to `ql` instead of `icwmrdp` as in the snapshot? I was not totally sure that `icwmrdp` corresponds exactly to `ql` since there is a `icwmrsh` which is the shallow convection equivalent, but I couldn't find a clear definition of the `ql` in the current CAM's state to be sure.

src/data/registry.xml
```diff
@@ -491,5 +491,76 @@
...
+              access="protected" >
+      <initial_value>.true.</initial_value>
+    </variable>
+    <variable local_name="il1g"
+              standard_name="index_of_first_column_of_gathered_deep_convection_arrays"
+              units="index" type="integer"
+              access="protected" >
+      <initial_value>1</initial_value>
+    </variable>
+    <variable local_name="dpdry"
```
jimmielin, line 551: Same question about the choice of the local name `dpdry` here (which usually implies contiguous, non-gathered dry delta-p layer thickness?) It appears the CCPP standard names spreadsheet calls this `fake_dpdry`. Since this is convective gathered columns maybe the naming could be clear that this is related to convection.

src/physics/utils/phys_comp.F90
```diff
@@ -177,12 +177,13 @@ subroutine phys_init()
...
       use physics_types,        only: allocate_physics_types_fields
       use cam_ccpp_cap,         only: cam_ccpp_physics_initialize
+      use cam_constituents,     only: num_advected
 
       call cam_thermo_init(columns_on_task, pver, pverp)
       call cam_thermo_formula_init()
 
       call allocate_physics_types_fields(columns_on_task, pver, pverp,        &
-           set_init_val_in=.true., reallocate_in=.false.)
+           num_advected, set_init_val_in=.true., reallocate_in=.false.)
```
jimmielin, line 186: Can I ask if this change is still necessary? Since `physics_types` is generated I wasn't sure if a framework change is necessary to accompany this change.

jimmielin (changes requested):
Thanks @cacraigucar! I had a couple questions on the `physics_types` naming and another regarding a subroutine call change, hopefully easy to address.

src/data/registry.xml
cacraigucar: The simple answer is that that is what the initial developer (whoever that was) used it in their code.  Inside zm_convr_intr.F90 which resides in ESCOMP/CAM, there is a pbuf call `call pbuf_get_field(pbuf, icwmrdp_idx,     ql )`.  That links the two variable names together.

src/data/registry.xml
cacraigucar: Again, this name is from the original developer and comes from ESCOMP/CAM zm_conv_intr.F90.  The `dpdry` name is used both as a value which has meaning and a `fake_dpdry`.  The time that dpdry has a value, it gets it from: `dpdry(i,:) = state%pdeldry(ideep(i),:)/100._r8`

src/physics/utils/phys_comp.F90
cacraigucar: I believe this code change was given to me by @peverwhee or @nusbaume.  I tried taking it out and I get the following compilation error: `Error: Missing actual argument for argument 'number_of_ccpp_constituents' at (1)`, so I deem it is necessary.  One of them can probably provide an explanation.

cacraigucar (commented):
I don't believe we should change either of the names that @jimmielin has concerns about since changes would need to be made is ESCOMP/CAM and ESCOMP/atmospheric_physics, though I am open to being overruled.

src/data/registry.xml
jimmielin: Thanks @cacraigucar. I don't think a change would need to be made in ESCOMP/CAM nor in the scheme to rename these in SIMA registry. 

The "link" of these variables to CAM snapshots is through `<ic_file_input_names>`. The link to the CCPP schemes themselves is through the standard name. 
I think the `local_name` here is only used for when these variables are `USE`d through the `physics_types` object.

To confirm this, I renamed `ql` and `dpdry` in `src/data/registry.xml` to `icwmrdp` and `dpdry_ideep` and built CAM-SIMA with ZM, and was able to get the same results as before the renaming - so I believe changing the `local_name` here won't affect anything else.

I am concerned that names that are too specific (in the case of ql/`icwmrdp` here, it is specific to deep convection: `_due_to_deep_convection`; for `dpdry`, it is for gathered columns: `_for_deep_convection_for_gathered_convective_columns`) are being given generic names (`ql` and `dpdry`) that otherwise have meanings outside of deep convection. The naming here (in the registry) applies across CAM-SIMA and not just within the ZM scheme, so I'd like to advocate for not using generic names but populating them with scheme-specific data.

Sorry for insisting on this, I'm happy to discuss further!

src/physics/utils/phys_comp.F90
peverwhee: hiya! `physics_types` is generated from the registry by CAM-SIMA rather than the CCPP framework. The new argument was added because Cheryl introduced a new dimension in the registry (number_of_ccpp_constituents / num_advected), which caused the generated code to expect this additional input

src/data/registry.xml
peverwhee: @jimmielin @cacraigucar

I lean towards @jimmielin 's approach here. These names are only used in the generated code in SIMA and the caps, so an update to atmospheric_physics wouldn't be necessary. And if there's a way to make them a little clearer so someone down the line doesn't think "hey, I need something called dpdry, i'll just use this standard name here in the registry", I think it'd be worth it.

src/physics/utils/phys_comp.F90
jimmielin: Thanks @cacraigucar and @peverwhee for explaining!

src/data/registry.xml
cacraigucar: Changed the two names as requested

src/data/registry.xml
cacraigucar: Changed as requested

jimmielin (approved):
Thanks @cacraigucar for addressing my comments!

cime_config/atm_in_paramgen.py
```diff
@@ -1495,11 +1495,21 @@ def write(self, output_path):
...
+                        emsg += "valid/default 'value' element."
+                        raise AtmInParamGenError(emsg)
+                    else:
+                        #If value found then strip white space:
+                        val = val.strip()
+                    #End if
+
+                    #If no value is set then move to the next variable:
+                    #if val is None:
+                    #    continue
```
peverwhee, line 1512: can these comments be removed?

src/data/registry.xml
```diff
@@ -491,5 +491,76 @@
...
+    <variable local_name="landfrac"
+              standard_name="land_area_fraction"
+              units="fraction" type="real" kind="kind_phys"
+              allocatable="allocatable">
+      <dimensions>horizontal_dimension</dimensions>
+      <ic_file_input_names>landfrac cam_in_landfrac</ic_file_input_names>
+    </variable>
+
+    <!-- Zhang McFarlane (ZM) Variables -->
+    <!-- Note that no_deep_pbl defaults to false, but needs to be set to true for diag_TKE or UW schemes -->
```
peverwhee, line 505: is this comment outdated? I don't see `no_deep_pbl` here.

src/physics/utils/ppgrid.F90
```diff
@@ -0,0 +1,24 @@
...
+!
+! Purpose:  USED ONLY FOR BACKWARDS COMPATIBILITY WITH CAM!!!!
+!           PLEASE DELETE ONCE NO LONGER NEEDED BY "to_be_ccppized"
+!           PHYSICS SCHEMES!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+!           Thanks!
+!-----------------------------------------------------------------------
+
+  use physics_grid,     only: pcols => columns_on_task
+  use vert_coord,       only: pver => pver, pverp => pverp !WILL NEED TO CHANGE THESE NAMES TO
+                                                           !SOMETHING ELSE IN CAM-SIMA!
```
peverwhee, line 14: i'm guessing this is a question for @nusbaume - do these variable names need to change?

peverwhee (approved):
a couple little cleanup things but nothing that is preventing my approval!

cime_config/testdefs/testlist_cam.xml
```diff
@@ -43,6 +43,15 @@
       <option name="comment">Test for Moist Held-Suarez (TJ2016) physics (physics_after_coupler group)</option>
     </options>
   </test>
+  <test compset="FPHYStest" grid="ne3pg3_ne3pg3_mg37" name="SMS_Ln2" testmods="cam/outfrq_zm_derecho">
```
peverwhee, line 46: oops! missed this during my review, but I believe you need to commit the new outfrq_zm_derecho directory.

src/data/registry.xml
```diff
@@ -491,5 +491,76 @@
...
+              standard_name="atmosphere_boundary_layer_thickness"
+              units="m" type="real" kind="kind_phys"
+              allocatable="allocatable">
+      <dimensions>horizontal_dimension</dimensions>
+      <ic_file_input_names>pblh pbuf_pblh</ic_file_input_names>
+    </variable>
+    <variable local_name="delt"
+              standard_name="half_timestep_for_physics"
+              units="s" type="real" kind="kind_phys">
+    </variable>
```
nusbaume, line 516: Is this variable needed anymore?  I couldn't see any use of it in the CCPP-ized ZM schemes.

cime_config/atm_in_paramgen.py
```diff
@@ -1495,11 +1495,21 @@ def write(self, output_path):
                 # Write all variables within that group (sorted alphabetically):
                 for var in sorted(self._data[nml_group], key=var_sort_key):
                     #Extract variable value(s):
-                    val = self._data[nml_group][var]["values"].strip()
+                    val = self._data[nml_group][var]["values"] #.strip()
```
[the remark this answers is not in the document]
cacraigucar: removed

cime_config/testdefs/testlist_cam.xml
cacraigucar: Will need to get this from derecho once it is back up

src/data/registry.xml
```diff
@@ -491,5 +491,76 @@
...
+              units="flag" type="logical"
+              access="protected" >
+      <initial_value>.true.</initial_value>
+    </variable>
+    <variable local_name="il1g"
+              standard_name="index_of_first_column_of_gathered_deep_convection_arrays"
+              units="index" type="integer"
+              access="protected" >
+      <initial_value>1</initial_value>
+    </variable>
```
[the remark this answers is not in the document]
nusbaume: Note the relevant issue can be found here:
https://github.com/ESCOMP/atmospheric_physics/issues/202

cime_config/testdefs/testlist_cam.xml
cacraigucar: Added

nusbaume (approved):
Looks great to me now, thanks @cacraigucar!

PR #358, opened 2025-02-06
Configure option syntax now only uses dashes

### LINKED ISSUE #357: Should convert CAM_CONFIG_OPTS argument syntax to use dashes instead of underscores
### Issue Type

Other (please describe below)

### Issue Description

Currently many of the arguments into `CAM_CONFIG_OPTS` in CAM-SIMA, although styled like command line inputs, use underscores instead of dashes (e.g. `--dyn_kind` instead of `--dyn-kind`).  Thus these arguments should be updated to use dashes instead, which is likely to be what the average user or developer would expect. 

### Will this change answers?

No

### Will you be implementing this yourself?

Yes

nusbaume:
Originator(s): nusbaume

Fixes #357 - Should convert CAM_CONFIG_OPTS argument syntax to use dashes instead of underscores

All arguments that can be used in the `CAM_CONFIG_OPTS` XML variable in a CAM-SIMA case's `env_build.xml` file now use dashes instead of underscores.  Also all single-dash config options (e.g. `-dyn` versus `--dyn`) have been removed to avoid any confusion as users move from CAM to CAM-SIMA). 

M       cime_config/cam_config.py
M       cime_config/config_component.xml
  - Convert CAM_CONFIG_OPTS argument syntax to use dashes instead of underscores, and remove single-dash options.  Also add an extra comment.

M       cime_config/testdefs/testmods_dirs/cam/outfrq_se_cslam_analy_ic/shell_commands
  - Update syntax for regression test that used analytic ICs to now use dashes (`--analytic-ic`)

M       test/unit/test_cam_config.py
  - Fix unit test structure that was using a no-longer-supported single-dash config argument.

derecho/intel/aux_sima:  ALL PASS

derecho/gnu/aux_sima:  ALL PASS

### DIFF
--- a/cime_config/cam_config.py
+++ b/cime_config/cam_config.py
@@ -589,7 +589,7 @@ class ConfigCAM:
         [('analytic_ic', False), ('dyn', 'se'), ('dyn_kind', 'REAL64'), ('phys_kind', 'REAL64'), ('physics_suites', 'kessler')]
 
         5.  Check that parse_config_opts works as expected when given both a string and logical argument:
-        >>> config_opts = ConfigCAM.parse_config_opts("--physics-suites kessler --dyn se --analytic_ic")
+        >>> config_opts = ConfigCAM.parse_config_opts("--physics-suites kessler --dyn se --analytic-ic")
         >>> vargs = vars(config_opts)
         >>> [(x, vargs[x]) for x in sorted(vargs)]
         [('analytic_ic', True), ('dyn', 'se'), ('dyn_kind', 'REAL64'), ('phys_kind', 'REAL64'), ('physics_suites', 'kessler')]
--- a/cime_config/cam_config.py
+++ b/cime_config/cam_config.py
@@ -613,26 +613,38 @@ class ConfigCAM:
                                          epilog="Allowed values of "+cco_str)
 
         #Add argument options:
-        parser.add_argument("--physics-suites", "-physics-suites", type=str,
-                            required=True, metavar='<CCPP_SDFs>',
+        parser.add_argument("--physics-suites",
+                            type=str,
+                            required=True,
+                            metavar='<CCPP_SDFs>',
                             help="""Semicolon-separated list of Physics Suite
                                  Definition Files (SDFs)""")
 
-        parser.add_argument("--dyn", "-dyn", metavar='<dycore>',
-                            type=str, required=False, default="",
+        parser.add_argument("--dyn",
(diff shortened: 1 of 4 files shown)


nusbaume:
Also wanted to note that I have created a new issue in the CAM-SIMA-docs repo (ESCOMP/CAM-SIMA-docs#6) that indicates the need to add some documentation that lists all of the config options, what they do, and what they map to in the actual python code.  I will add that page/section once this PR is merged.  Thanks!

cime_config/cam_config.py
```diff
@@ -618,16 +622,16 @@ def parse_config_opts(cls, config_opts, test_mode=False):
                             type=str, required=False, default="",
                             help="""Name of dycore""")
 
-        parser.add_argument("--analytic_ic", "-analytic_ic",
+        parser.add_argument("--analytic-ic", "-analytic-ic",
```
jimmielin, line 625: Would we still want to support the single-dash versions (e.g., `-analytic-ic`) that's similar to current CAM if it has changed (from `-analytic_ic` to `-analytic-ic`) anyway?

jimmielin (approved):
Thanks @nusbaume! I just had one question about the single-dash variant, otherwise it looks great!

boulderdaze (approved):
Looks good, thank you!

cime_config/cam_config.py
nusbaume: I personally think having just the one dash as an option is ok in case someone typos when manually adding the argument, but if you disagree then I am happy to take it out.  Just let me know!

boulderdaze: Yeah I wonder this too.

jimmielin: Thanks @nusbaume! Personally I would be in favor of getting rid of it to have less 'historical' stuff in SIMA that behaves kinda like CAM but not exactly like it (it would be surprising for the user, for example, for `-chem something` to work like CAM in the future but `-analytic_ic` to not work). But I am happy either way.

boulderdaze: I am not familiar with CAM. I only mentioned that in the context of the command line interface where you expect with double dash, more descriptive flags, and with a single dash, a quick, shorthand command. (not always though). I didn't know it's from CAM

boulderdaze: I like it, thank you!

jimmielin (approved):
Thank you!

Issue #359, opened 2025-02-11
Setting <initial_value> in registry should allow model to run without reading field from ic file


jimmielin:
### What happened?

Add a registry.xml variable that can be read from the snapshot, but ordinarily does not exist in an initial condition file (or provided by analytic_ic)
```xml
    <variable local_name="cmfmc"
              standard_name="atmosphere_convective_mass_flux_due_to_all_convection"
              units="kg m-2 s-1" type="real" kind="kind_phys"
              allocatable="allocatable">
      <dimensions>horizontal_dimension vertical_interface_dimension</dimensions>
      <initial_value>0.0_kind_phys</initial_value>
      <ic_file_input_names>cmfmc tphysbc_cmfmc</ic_file_input_names>
    </variable>
```

Run the model with dycore and analytic ic condition. The model crashes
```
 ERROR: read_field_3d: No variable found in (/ cmfmc, tphysbc_cmfmc,  /)
```

These fields are generally "snapshot" fields (e.g., `tphysbc` or `pbuf` fields) that are part of the model state and thus necessary for snapshot testing, but not necessary for a full run.

This should probably be changed to:
1. if variable is found in ic file (e.g., snapshot) or analytic ic - read it in
2. if variable is not found, allow the model to continue running while setting this variable to the `initial_value` (possibly show a warning -- eventually the warning can be suppressed once registry.xml is split into "regular" and "snapshot-only")

The guideline for adding `registry.xml` fields should probably be: for "regular" fields, don't specify an `<initial_value>` (it should always be provided by initial conditions); otherwise, allow initial condition files to not have that variable available.

### Analysis

This happens despite generated code, `physics_types.F90`, assigns the initial value correctly
```fortran
      if (allocated(cmfmc)) then
         if (reallocate) then
            deallocate(cmfmc)
         else
            call endrun(subname//": cmfmc is already allocated, cannot allocate")
         end if
      end if
      allocate(cmfmc(horizontal_dimension, vertical_interface_dimension))
      if (set_init_val) then
         cmfmc = 0.0_kind_phys
      end if
```

`physics_inputs.F90` still attempts to call `check_field` and read this field, despite the initial_value provided.

### What are the steps to reproduce the bug?

See above.

### What CAM-SIMA hash were you using?

head

### What machine were you running CAM-SIMA on?

CISL machine (e.g. cheyenne)

### What compiler were you using?

Intel

### Will you be addressing this bug yourself?

Any Software Engineer can do this

PR #362, opened 2025-02-13
Complete CCPPIzation of Hack shallow convection

jimmielin:
Originator(s): jimmielin 

- Completes CCPPization of Hack shallow convection.
- Adds FCAM4 (CAM4 physics compset) and CAM4 tests.

```
M       .gitmodules
M       src/physics/ncar_ccpp
  - bring in updated atmospheric physics with Hack shallow convection.
  - bring in updated atmospheric physics with suite_cam4.xml.

M       cime_config/config_component.xml
M       cime_config/config_compsets.xml
M       cime_config/testdefs/testlist_cam.xml
  - add CAM4 suite and test

M       src/data/registry.xml
  - registry updates for reading necessary fields for Hack shallow convection snapshot test
  - registry updates for use_shfrc and shfrc for future cloud_fraction scheme (persistent)
```

derecho/intel/aux_sima: 

derecho/gnu/aux_sima:


jimmielin:
Holding as draft until I finish testing and documenting that the CAM4 tests will likely not run because all the snapshot variables are missing and model would crash (#359)

peverwhee:
> Updated to `atmos_phys0_10_001`! All tests pass/new baselines (there appears to be some NLCOMP failures due to external updates?) and ready for review.

Hmmm... Maybe you need to rerun git-fleximod? Or get a clean checkout? Peeking at your tests, it looks like they match the old (pre-externals-update) baselines. Let me know if it still isn't working though. Could  be a me issue for sure!

peverwhee:
@jimmielin oh. Looking at the git log of your local clone on derecho, your branch isn't at the head of CAM-SIMA development. Once you merge up, those NLCOMP failures should go away. Apologies if you already knew all of this and I'm just catching up.

src/data/registry.xml
```diff
@@ -498,24 +581,87 @@
...
+      <dimensions>horizontal_dimension vertical_layer_dimension</dimensions>
+      <!-- only used if use_shfrc is .true. -->
+      <initial_value>0.0_kind_phys</initial_value>
+    </variable>
+
+    <!-- Zhang McFarlane (ZM) Variables -->
+    <variable local_name="delt"
+              standard_name="half_timestep_for_physics"
+              units="s" type="real" kind="kind_phys">
+    </variable>
```
nusbaume, line 664: Is this needed?  I don't think ZM actually uses a half-physics timestep.

nusbaume (approved):
Thanks @jimmielin!  I just have one small request (but I don't think it needs a re-review).

src/data/registry.xml
```diff
@@ -498,24 +581,83 @@
...
     <variable local_name="tpert"
               standard_name="convective_temperature_perturbation_due_to_pbl_eddies"
               units="K" type="real" kind="kind_phys"
               allocatable="allocatable">
       <dimensions>horizontal_dimension</dimensions>
+      <initial_value>0.0_kind_phys</initial_value>
       <ic_file_input_names>tpert pbuf_tpert</ic_file_input_names>
     </variable>
+
+    <!-- Cloud fraction variables - need to be kept in physics state -->
```
peverwhee, line 635: Perhaps a naive comment, but what does "need to be kept in physics state" mean?
1. If it means it should be part of the physics state DDT, then it (and convect_shallow_use_shfrc) should be added to that DDT around line~399. 
2. If it means it should be kept in the registry (in contrast to the temporary variables), then maybe a verbiage change here and below to "need to be kept in the CAM registry"?
3. If it means some third thing... what is it? :)

src/data/registry.xml
```diff
@@ -137,6 +137,89 @@
...
+    <variable local_name="rprddp"
+              standard_name="tendency_of_precipitation_wrt_moist_air_and_condensed_water_due_to_deep_convection_excluding_subcloud_evaporation"
+              units="kg kg-1 s-1" type="real" kind="kind_phys"
+              allocatable="allocatable">
+      <dimensions>horizontal_dimension vertical_layer_dimension</dimensions>
+      <initial_value>0.0_kind_phys</initial_value>
+      <ic_file_input_names>pbuf_RPRDDP</ic_file_input_names>
+    </variable>
+    <!-- note: rliq and rliq2 from CAM snapshot appear to be in m s-1 (precip) units.
+         the CCPP standard names table claim that it should be in kg m-2 s-1 units. -->
```
peverwhee, line 206: do we need to look into this? e.g. make an issue somewhere?

peverwhee (approved):
Thanks @jimmielin I have a couple of questions in the registry, but nothing that prevents my approval!

src/data/registry.xml
peverwhee: Ah, then I think updating it to refer to the registry rather than the physics state would be best. The physics_state DDT is for quantities that define the "state" of the atmosphere (like temperature, pressure, etc.). 

Also, the generated code from the registry is in `physics_types` rather than physics_state (which is a DDT within physics_types. So maybe what you meant to begin with was "kept in physics types"? If so, that would work for me as well.

peverwhee: And I have just reread your question and feel like I didn't quite answer it.

I believe the physics_state DDT primarily exists for the dycore since that expects a state DDT, and then the framework handles the physics interfaces (which don't really care where the data is coming from - DDT or otherwise)

peverwhee: Thanks @jimmielin !

PR #365, opened 2025-02-20
Allow registry fields with initial_value to not error on not found in input data

### LINKED ISSUE #359: Setting <initial_value> in registry should allow model to run without reading field from ic file

jimmielin:
Originator(s): @jimmielin (with extensive guidance from @peverwhee)

* Fixes #359 

Should enable CAM7 / CAM4 regression tests to run again despite no PBL scheme, if `pbuf_pblh` includes `<initial_value>0.0_kind_phys</initial_value>` in registry.xml - separate PR will include this.

* Adds `vars_init_value` in generation of physics_inputs.F90 to record which variables have an `<initial_value>` set in `registry.xml`

```
A       test/unit/python/sample_files/write_init_files/phys_vars_init_check_initial_value.F90
A       test/unit/python/sample_files/write_init_files/physics_inputs_initial_value.F90
A       test/unit/python/sample_files/write_init_files/simple_reg_initial_value.xml
  - new test sample files for write_init_files
```

```
M       cime_config/cam_autogen.py
M       cime_config/cam_build_cache.py
M       cime_config/cam_config.py
M       src/data/generate_registry_data.py
  - pass variables having initial_value element in registry.xml

M       src/data/write_init_files.py
  - add error_on_not_found=.false. to variables having initial_value element
  - or write info message for variables that cannot be read (i.e., no horizontal dimension)

M       test/unit/python/sample_files/build_cache_files/update_reg_build_cache.xml
M       test/unit/python/test_build_cache.py
M       test/unit/python/test_cam_autogen.py
M       test/unit/python/test_registry.py
M       test/unit/python/test_write_init_files.py
  - fix compatibility of existing tests with new vars_init_value argument
  - add test_simple_initial_value_write_init unit test
```

derecho/intel/aux_sima:

derecho/gnu/aux_sima:

### DIFF
--- a/src/data/write_init_files.py
+++ b/src/data/write_init_files.py
@@ -48,8 +48,7 @@ _MAX_LINE_LEN = 200
 ##############
 
 def write_init_files(cap_database, ic_names, registry_constituents, vars_init_value,
-                     outdir,
-                     file_find_func, source_paths, indent, logger,
+                     outdir, file_find_func, source_paths, indent, logger,
                      phys_check_filename=None, phys_input_filename=None):
 
     """
--- a/test/unit/python/test_write_init_files.py
+++ b/test/unit/python/test_write_init_files.py
@@ -341,7 +341,7 @@ class WriteInitTest(unittest.TestCase):
         cap_database = capgen(run_env, return_db=True)
 
         # Generate physics initialization files:
-        retmsg = write_init.write_init_files(cap_database, ic_names, [], ["air_pressure_at_sea_level"], _TMP_DIR,
+        retmsg = write_init.write_init_files(cap_database, ic_names, [], vars_init_value, _TMP_DIR,
                                              find_file, _INC_SEARCH_DIRS,
                                              3, logger,
                                              phys_check_filename=vic_name,


test/unit/python/test_write_init_files.py
```diff
@@ -238,10 +238,10 @@ def test_simple_reg_constituent_write_init(self):
                       check_init_out, phys_input_out])
 
         # Generate registry files:
-        _, _, ic_names, constituents = gen_registry(filename, 'se', _TMP_DIR, 3,
-                                      _SRC_MOD_DIR, _CAM_ROOT,
-                                      loglevel=logging.ERROR,
-                                      error_on_no_validate=True)
+        _, _, ic_names, constituents, _ = gen_registry(filename, 'se', _TMP_DIR, 3,
```
peverwhee, line 241: I think it would be good to add a new test (mirroring this one `test_simple_reg_constituent_write_init`, perhaps) that tests write_init_files.py when there is an vars_init_value list (to make sure the calls are added correctly)

peverwhee (changes requested):
thanks @jimmielin ! 

One testing request, but otherwise no notes! Excellent work on the somewhat confusing build cache updates.

peverwhee (approved):
awesome thanks again @jimmielin

test/unit/python/test_write_init_files.py
```diff
@@ -258,7 +258,90 @@ def test_simple_reg_constituent_write_init(self):
...
+                                   host_name='cam', kind_types=kind_types,
+                                   use_error_obj=False,
+                                   force_overwrite=True,
+                                   output_root=_TMP_DIR,
+                                   ccpp_datafile=cap_datafile)
+
+        cap_database = capgen(run_env, return_db=True)
+
+        # Generate physics initialization files:
+        retmsg = write_init.write_init_files(cap_database, ic_names, [], ["air_pressure_at_sea_level"], _TMP_DIR,
```
nusbaume, line 344: Instead of hard-coding `air_pressure_at_sea_level` here I would use the `vars_init_value` passed out by `gen_registry`:

```suggestion
        retmsg = write_init.write_init_files(cap_database, ic_names, [], vars_init_value, _TMP_DIR,
```

src/data/write_init_files.py
```diff
@@ -47,7 +47,8 @@
 #Main function
 ##############
 
-def write_init_files(cap_database, ic_names, registry_constituents, outdir,
+def write_init_files(cap_database, ic_names, registry_constituents, vars_init_value,
+                     outdir,
                      file_find_func, source_paths, indent, logger,
                      phys_check_filename=None, phys_input_filename=None):
```
nusbaume, line 53: The spacing looks a little strang here, maybe do something like this instead (?):

```suggestion
def write_init_files(cap_database, ic_names, registry_constituents, vars_init_value,
                     outdir, file_find_func, source_paths, indent, logger,
                     phys_check_filename=None, phys_input_filename=None):
```

nusbaume (approved):
Thanks @jimmielin!  Just had two minor requests (but they don't require a full re-review).

PR #386, opened 2025-04-24
Check for and add missing error code handling for PIO calls

### LINKED ISSUE #175: Check PIO return codes
Many PIO interfaces are functions which return an error code (in `PIO_BCAST_ERROR` mode). CAMDEN code should be more diligent about using `cam_pio_handle_error` to check these codes where a failure should be fatal.

cacraigucar:
Originator(s): cacraig

Closes #175: Check PIO return codes

  - src/dynamics/none/dyn_grid.F90---------     No PIO function return codes found that weren't checked
- src/dynamics/utils/hycoef.F90------------ No PIO function return codes found that weren't checked
- src/dynamics/se/native_mapping.F90---- No PIO function return codes found that weren't checked
- src/dynamics/se/dyn_comp.F90---------- Put in one check (was probably unnecessary, but won't hurt)
- src/dynamics/se/dyn_grid.F90------------ No PIO function return codes found that weren't checked
- src/history/cam_history_support.F90----- No PIO function return codes found that weren't checked
- src/cpl/nuopc/atm_comp_nuopc.F90----- No PIO function return codes found that weren't checked
- src/cpl/nuopc/atm_stream_ndep.F90-----  No PIO function return codes found that weren't checked
- src/utils/cam_pio_utils.F90----------------    Added a comment before one block which did not check, but it was doing internal PIO checking
- src/utils/cam_grid_support.F90----------- No PIO function return codes found that weren't checked
- src/utils/cam_time_coord.F90-------------    No PIO function return codes found that weren't checked
- src/physics/utils/solar_parms_data.F90--- No PIO function return codes found that weren't checked

M       src/dynamics/se/dyn_comp.F90
             - added a check, which was probably unnecessary, but does no harm
             
M       src/utils/cam_pio_utils.F90
              - added comment before a block of code which is not being checked as PIO is doing internal checking

derecho/intel/aux_sima:

derecho/gnu/aux_sima:

### DIFF
--- a/src/utils/cam_pio_utils.F90
+++ b/src/utils/cam_pio_utils.F90
@@ -940,6 +940,7 @@ CONTAINS
             strt = 1     ! cnt set by cam_pio_var_info
             exists = use_scam_limits(File, strt, cnt,filedims)
          end if
+         ! Internal PIO error handling is in effect, so need to check return codes
          if (ndims == 3) then
             ierr = pio_inq_dimname(File, dimids(3), tmpname)
             if (trim(tmpname) /= 'time') then
--- a/src/utils/cam_pio_utils.F90
+++ b/src/utils/cam_pio_utils.F90
@@ -940,7 +940,7 @@ CONTAINS
             strt = 1     ! cnt set by cam_pio_var_info
             exists = use_scam_limits(File, strt, cnt,filedims)
          end if
-         ! Internal PIO error handling is in effect, so need to check return codes
+         ! Internal PIO error handling is in effect, so no need to check return codes
          if (ndims == 3) then
             ierr = pio_inq_dimname(File, dimids(3), tmpname)
             if (trim(tmpname) /= 'time') then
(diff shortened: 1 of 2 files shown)


src/utils/cam_pio_utils.F90
```diff
@@ -940,6 +940,7 @@ subroutine cam_pio_get_var_2d_r8(varname, File, field, start, kount, found)
             strt = 1     ! cnt set by cam_pio_var_info
             exists = use_scam_limits(File, strt, cnt,filedims)
          end if
+         ! Internal PIO error handling is in effect, so need to check return codes
```
nusbaume, line 943: Did you mean to say that there is no need here?

```suggestion
         ! Internal PIO error handling is in effect, so no need to check return codes
```

nusbaume (changes requested):
Thanks @cacraigucar!  Had one comment change request, but otherwise everything looks good to me.

src/utils/cam_pio_utils.F90
cacraigucar: Oops - fixed!

peverwhee (approved):
thanks @cacraigucar !

PR #394, opened 2025-05-23
Enable MPAS dynamical core to compute relative vorticities at cell points

kuanchihwang:
### Tag name (required for release branches):

None

### Originator(s):

kuanchihwang

### Descriptions (include the issue title, and the keyword ['closes', 'fixes', 'resolves'] followed by the issue number):

This PR ports a new feature from ESCOMP/CAM#1297 to CAM-SIMA. It enables MPAS dynamical core to compute relative vorticities at cell points. Although the feature is currently not used anywhere, it will be when CAM7 physics is implemented in CAM-SIMA.

The formulations are identical to CAM, but the implementation here incorporates object-oriented programming to be consistent with other parts of MPAS dynamical core.

### Describe any changes made to the build system:

None

### Describe any changes made to the namelist:

None

### List any changes to the defaults for the input datasets (e.g., boundary datasets):

None

### List all files eliminated and why:

None

### List all files added and what they do:

None

### List all existing files that have been modified, and describe the changes:

```
M       src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
  * Enable MPAS subdriver to compute relative vorticities at cell points
```

### Regression tests:

No changes to any existing tests. All tests pass with respect to the last baseline, `sima0_05_001`.

### DIFF
--- a/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
+++ b/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
@@ -110,6 +110,7 @@ module dyn_mpas_subdriver
         procedure, pass, public :: exchange_halo => dyn_mpas_exchange_halo
         procedure, pass, public :: compute_unit_vector => dyn_mpas_compute_unit_vector
         procedure, pass, public :: compute_edge_wind => dyn_mpas_compute_edge_wind
+        procedure, pass, public :: compute_cell_relative_vorticity => dyn_mpas_compute_cell_relative_vorticity
         procedure, pass, public :: init_phase4 => dyn_mpas_init_phase4
         procedure, pass, public :: run => dyn_mpas_run
         procedure, pass, public :: final => dyn_mpas_final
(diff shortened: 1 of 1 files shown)


src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -2776,6 +2777,106 @@ subroutine dyn_mpas_compute_edge_wind(self, wind_tendency)
...
+        call self % get_variable_pointer(kiteforcell, 'mesh', 'kiteForCell')
+        call self % get_variable_pointer(nedgesoncell, 'mesh', 'nEdgesOnCell')
+        call self % get_variable_pointer(verticesoncell, 'mesh', 'verticesOnCell')
+
+        call self % get_variable_pointer(areacell, 'mesh', 'areaCell')
+        call self % get_variable_pointer(kiteareasonvertex, 'mesh', 'kiteAreasOnVertex')
+        call self % get_variable_pointer(vorticity, 'diag', 'vorticity')
+
+        ! Output.
+        allocate(cell_relative_vorticity(nvertlevels, ncellssolve), stat=ierr)
```
nusbaume, line 2823: Just an FYI that we have started trying to provide the Fortran-provided error message from allocate statements as well, e.g.:

```
character(256) :: errmsg
...
allocate(cell_relative_vorticity(nvertlevels, ncellssolve), stat=ierr, errmsg=errmsg)
```

Of course this change certainly isn't required for this PR.

nusbaume (approved):
Thanks again @kuanchihwang!  Just had one general comment (but it doesn't require any changes to this particular PR).

src/dynamics/mpas/driver/dyn_mpas_subdriver.F90
```diff
@@ -2776,6 +2777,106 @@ subroutine dyn_mpas_compute_edge_wind(self, wind_tendency)
...
+            do k = 1, nvertlevels
+                cell_relative_vorticity(k, i) = regrid_from_vertex_to_cell(i, k, &
+                    nedgesoncell, verticesoncell, kiteforcell, kiteareasonvertex, areacell, &
+                    vorticity)
+            end do
+        end do
+
+        nullify(ncellssolve, nvertlevels)
+        nullify(kiteforcell, nedgesoncell, verticesoncell)
+        nullify(areacell, kiteareasonvertex, vorticity)
```
cacraigucar, line 2839: I didn't dive into these pointers to see if they are being intenally allocated or not.  If they are, then a deallocate statement should happen before the pointers are nullified.  I assume they are truly just pointers, in which case these lines are just fine as written.

kuanchihwang: @cacraigucar

Thanks for the question!

These pointers point to the variables that are allocated/deallocated by MPAS. If we prematurely deallocate them here, MPAS will actually lose access to them and will most likely crash.

If you look a little further up in this subroutine, you can find that these pointers become associated with their targets by calling `get_variable_pointer`. Its code comments clarify<sup>[[1]](https://github.com/ESCOMP/CAM-SIMA/blob/sima0_06_000/src/dynamics/mpas/driver/dyn_mpas_subdriver.F90#L3645-L3662)</sup> that it is "... for drawing the REFERENCE of an internal variable from MPAS dynamical core instance".

cacraigucar (approved):

PR #405, opened 2025-07-08
Use default value for initial chemical species mixing ratios

mattldawson:
Originator(s): Matt Dawson

This PR updates to use a newer version of `atmospheric_physics` that includes setting the `default_value` of constituent instances created for chemical species based on data in the MICM species configuration files. The updates to CAM-SIMA remove the hard-coded initial species mixing ratios for Terminator and Chapman mechanisms and instead set the initial mixing ratios for chemical species based on the `default_value` of the corresponding constituent, which allows any mechanism to be run. This is needed to demonstrate the run-time configurability of the chemical system.

updated commits for `atmospheric_physics` and `cam-sima-chemistry-data` repos.

none

none

`musica_sima_namelist.F90` - no namelist parameters are needed for musica now that CAM-SIMA is not restricted to Terminator and Chapman

will add

will add

will add after tests are run

derecho/intel/aux_sima:

derecho/gnu/aux_sima:

this should only affect chemistry


boulderdaze (approved):
Looks good!

nusbaume (approved):
Thanks @mattldawson!  I pushed up one last change to remove the `musica_config` namelist variable from the definition file (as the namelist reader for it was already removed), but otherwise it looks good to me! 

Also I think for now just setting the initial/default values for the chemical species only is fine, as some CAM physics schemes seem to require specialized initialization logic for their relevant constituent variables, and so until that is all sorted out it is probably safest to keep things separate.  Thanks for checking though!

PR #487, opened 2026-04-09
sima0_14_001: Add support for accumulated history fields

peverwhee:
Tag name (required for release branches): sima0_14_001
Originator(s): peverwhee
AI tools used (if applicable; please also add the "AI-generated code" label to the PR): n/a
  What:
  How:

* Adds support for accumulated (min, max, stdev [var], avg) fields
* Allows for field-specific fill values (passed into add field and populated in the hist_field object)
* Incorporates changes for, when write_nstep0=true for a file, the nstep0 file is written and then immediately closed ("real" run samples not included)
* Adds logic to clobber accumulated field settings (switch to instantaneous) if the output frequency for a file is 1*nsteps (every step) - same behavior as in CAM
* Adds logic to error during preview_namelists when two fields are on the same history file with different flags

```
M  cime_config/atm_in_paramgen.py
- small bugfix for namelist parsing
M  cime_config/hist_config.py
- clobber accumulated settings if frequency is every timestep
```  

List all files added and what they do: n/a

```
M  .gitmodules
M  src/history/buffers
- bring in latest history external

M  src/history/cam_hist_file.F90
- add new reset_samples function; clear the buffers after writing to file
- get norm value with real64 buffer, then set to real32 before writing (if applicable)
- add missing error handling

M  src/history/cam_history.F90
- don't clear the buffers for every file every timestep! obviously!
- pass fill_value to file object
- modify logic to handle nstep0 writing (to it's own file)
- add missing error handling
```

derecho/intel/aux_sima:
```
SMS_Ln9.mpasa120_mpasa120.FKESSLER.derecho_intel.cam-outfrq_kessler_mpas_derecho_history (Overall: DIFF)
SMS_Ln9.mpasa480_mpasa480.FKESSLER.derecho_intel.cam-outfrq_kessler_mpas_derecho (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FADIAB.derecho_intel.cam-outfrq_se_cslam (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FCAM7.derecho_intel.cam-outfrq_se_cslam_analy_ic (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FHS94.derecho_intel.cam-outfrq_se_cslam (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FKESSLER.derecho_intel.cam-outfrq_se_cslam (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FKESSLER.derecho_intel.cam-outfrq_se_cslam_multitape (Overall: DIFF) 
SMS_Ln9.ne3pg3_ne3pg3_mg37.FTJ16.derecho_intel.cam-outfrq_se_cslam (Overall: DIFF)
- all tests with write_nstep0=true are DIFF because of the new nstep0 behavior. Manually confirmed that answers have not changed.

SMS_Ln9.mpasa480_mpasa480.FKESSLER.derecho_intel.cam-outfrq_kessler_mpas_derecho_history (Overall: DIFF)
- new test; new baseline- 
```

derecho/gnu/aux_sima:
```
SMS_Ln2.ne3pg3_ne3pg3_mg37.FPHYStest.derecho_gnu.cam-outfrq_dme_adjust_derecho (Overall: DIFF) 
SMS_Ln2.ne3pg3_ne3pg3_mg37.FPHYStest.derecho_gnu.cam-outfrq_held_suarez_derecho (Overall: DIFF) 
SMS_Ln2.ne3pg3_ne3pg3_mg37.FPHYStest.derecho_gnu.cam-outfrq_kessler_derecho (Overall: DIFF) 
SMS_Ln2.ne3pg3_ne3pg3_mg37.FPHYStest.derecho_gnu.cam-outfrq_tj2016_after_coupler_derecho (Overall: DIFF)
SMS_Ln2.ne3pg3_ne3pg3_mg37.FPHYStest.derecho_gnu.cam-outfrq_tj2016_before_coupler_derecho (Overall: DIFF)
SMS_Ln9.mpasa480_mpasa480.FKESSLER.derecho_gnu.cam-outfrq_kessler_mpas_derecho (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FADIAB.derecho_gnu.cam-outfrq_se_cslam (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FCAM7.derecho_gnu.cam-outfrq_se_cslam_analy_ic (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FHS94.derecho_gnu.cam-outfrq_se_cslam (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FKESSLER.derecho_gnu.cam-outfrq_se_cslam (Overall: DIFF)
SMS_Ln9.ne3pg3_ne3pg3_mg37.FTJ16.derecho_gnu.cam-outfrq_se_cslam (Overall: DIFF)
- all tests with write_nstep0=true are DIFF because of the new nstep0 behavior. Manually confirmed that answers have not changed.

SMS_Ln9.mpasa480_mpasa480.FKESSLER.derecho_gnu.cam-outfrq_kessler_mpas_derecho_history (Overall: DIFF)
- new test; new baseline
```

derecho/nvhpc/aux_sima: run via github workflow


cime_config/hist_config.py
```diff
@@ -409,6 +409,9 @@ def num_fields(self):
         """Return the number of fields in this HistFieldList object."""
         return len(self.__field_names)
 
+    @property
+    def field_names(self):
+        return self.__field_names
```
jimmielin: Just a heads up that this looks like this is a duplicate of https://github.com/ESCOMP/CAM-SIMA/blob/ad5b5a5a49711ce4b83c4b0127266835d6448eca/cime_config/hist_config.py#L476-L479 further down in the module.

I did have another question on this - this might be my wrong understanding of Python - and if it is better to return a copy (`return list(self.__field_names)` in `field_names()`). I am worried that the `remove_fields` calls added below in `convert_accumulated_fields`, e.g.,
`self.__var_fields.remove_fields(self.__var_fields.field_names, pobj, logger)`

looking at the definition of `HistFieldList.remove_fields`
https://github.com/ESCOMP/CAM-SIMA/blob/ad5b5a5a49711ce4b83c4b0127266835d6448eca/cime_config/hist_config.py#L388-L396
it is iterating on `for field_name in fields` - where `fields` is the input argument `self.__var_fields.field_names`: right now `return self.__field_names`, which (I think?) is shallow copy of `self.__var_fields.__field_names`

As elements are removed from `self.__field_names`, the array itself changes, so `fields` will change too. I don't know if this might cause the `remove_fields` operation to skip every other element (since it is iterating on a list that has its elements being removed at the same time), but it would be good to check the `removed_fields` return value to be sure.

peverwhee: @jimmielin good call on the duplicate property! removing that now.

as for the field removal, i was worried about that as well, but had convinced myself it was behaving correctly. BUT i just ran a test and it did, in fact, skip the second field it was removing. I'll fix that too!

jimmielin: Thanks @peverwhee! Excited to use it to test the BAM optics branch I'm working on 😄

I am never sure of how lists are copied or not in Python but figured I might ask just in case!

peverwhee: thanks again @jimmielin - should be fixed now.

cime_config/testdefs/testlist_cam.xml
```diff
@@ -177,6 +177,18 @@
     </options>
   </test>
 
+  <!-- History output test -->
+  <test compset="FKESSLER" grid="mpasa120_mpasa120" name="SMS_Ln9" testmods="cam/outfrq_kessler_mpas_derecho_history">
```
nusbaume: I believe the resolution of this test can be dropped for the `FKESSLER` configuration (which should keep the test on the cheaper-side):


```suggestion
  <test compset="FKESSLER" grid="mpasa480_mpasa480" name="SMS_Ln9" testmods="cam/outfrq_kessler_mpas_derecho_history">
```

cime_config/testdefs/testmods_dirs/cam/outfrq_kessler_mpas_derecho_history/user_nl_cam
```diff
@@ -0,0 +1,8 @@
+debug_output=0
+hist_add_inst_fields;h3: Q,T,U,V,PS
+hist_add_avg_fields;h3: PSDRY
+hist_add_max_fields;h3: PHIS
+hist_add_min_fields;h3: PDEL, PDELDRY
+hist_add_var_fields;h3: ZM, PRECT
```
nusbaume, line 6: What happens if a variable is listed as both, say, an `avg` field and a `max` field?  For example:

```
hist_add_avg_fields;h3: PSDRY
hist_add_max_fields;h3: PSDRY
```

Is this gracefully handled in CAM and/or SIMA?  If not, should we add some sort of error checking so that users know that won't work at `preview_namelists` time?  We should probably update or add a test as well to ensure everything is working as expected (either writing to the file correctly or throwing a helpful error).

src/history/cam_hist_file.F90
```diff
@@ -762,27 +763,25 @@ subroutine config_set_up_fields(this, possible_field_list)
...
+            field_ptr%fill_value(), beg_dims=beg_dim,            &
+            end_dims=end_dim, mixing_ratio=field_ptr%mixing_ratio(), &
             flag_xyfill=field_ptr%flag_xyfill(),                 &
             sampling_seq=field_ptr%sampling_sequence())
          call hist_new_buffer(field_info, field_shape, &
-            this%rl_kind, 1, this%accumulate_types(idx), 1, errors=errors)
-         if (masterproc) then
+            1, this%accumulate_types(idx), 1, errors=errors)
+         if (masterproc .and. errors%num_errors() > 0) then
             call errors%output(iulog)
```
nusbaume, line 782: Can we open an issue in the history_output repo to see if we can get the error output passed to a string variable as well?  That would help situations like this where it would be helpful to include the error message in the `endrun` string as well.

I should also note that this is not a high-priority issue, just a "nice-to-have".

src/history/cam_hist_file.F90
```diff
@@ -1558,6 +1558,8 @@ subroutine config_write_field(this, field, split_file_index, restart, &
          if (trim(field_precision) == 'REAL32') then
             allocate(field_data_r4(end_dims(1) - beg_dims(1) + 1, 1), stat=ierr, errmsg=errmsg)
             call check_allocate(ierr, subname, 'field_data_r4', file=__FILE__, line=__LINE__-1, errmsg=errmsg)
+            allocate(field_data_r8(end_dims(1) - beg_dims(1) + 1, 1), stat=ierr, errmsg=errmsg)
+            call check_allocate(ierr, subname, 'field_data_r8', file=__FILE__, line=__LINE__-1, errmsg=errmsg)
```
nusbaume: Do we need this extra variable and precision-checking logic at all, or can we just do the kind conversion in the call to `cam_grid_write_dist_array`?  For example we could just have one `field_data_r8` variable that is always allocated, and then do `real(field_data_r8, r4)` when needed (see next comment).

src/history/cam_hist_file.F90
```diff
@@ -1586,30 +1590,42 @@ subroutine config_write_field(this, field, split_file_index, restart, &
...
+            if (errors%num_errors() > 0) then
+               call errors%output(iulog)
+               write(errmsg, *) subname, 'ERROR writing field "', trim(field%diag_name()), '"'
+               call endrun(errmsg)
+            end if
             if (trim(field_precision) == 'REAL32') then
-               call hist_buffer_norm_value(buff_ptr, field_data_r4(:,1))
+               field_data_r4(:,1) = field_data_r8(:,1)
                call cam_grid_write_dist_array(this%hist_files(split_file_index), field_decomp, (/dim_sizes(1)/), &
                     field_shape, field_data_r4(:,1), varid)
```
nusbaume: If you just do the single `field_data_r8` allocation as I suggested above, then this call becomes:


```suggestion
call cam_grid_write_dist_array(this%hist_files(split_file_index), field_decomp, (/dim_sizes(1)/), &
field_shape, real(field_data_r8(:,1), r4), varid)
```

src/history/cam_hist_file.F90
```diff
@@ -1586,30 +1590,42 @@ subroutine config_write_field(this, field, split_file_index, restart, &
...
+            if (errors%num_errors() > 0) then
+               call errors%output(iulog)
+               write(errmsg, *) subname, 'ERROR writing field "', trim(field%diag_name()), '"'
+               call endrun(errmsg)
+            end if
             if (trim(field_precision) == 'REAL32') then
-               call hist_buffer_norm_value(buff_ptr, field_data_r4)
+               field_data_r4 = field_data_r8
                call cam_grid_write_dist_array(this%hist_files(split_file_index), field_decomp, dim_sizes(1:frank), &
                     field_shape, field_data_r4, varid)
```
nusbaume: Just noting that if you accept my "single r8 variable" suggestion then this becomes:


```suggestion
call cam_grid_write_dist_array(this%hist_files(split_file_index), field_decomp, dim_sizes(1:frank), &
field_shape, real(field_data_r8, r4), varid)
```

src/history/cam_history.F90
```diff
@@ -376,9 +382,10 @@ end subroutine set_up_field_list_hash_table
 !===========================================================================
 
    subroutine history_add_field_1d(diagnostic_name, standard_name, vdim_name, &
```
nusbaume, line 384: I wasn't sure where to add this to your PR, but we probably want to open an issue to eventually add `r4` versions of `history_out_field` that then explicitly convert the `r4` inputs into `r8` before passing it into the buffer, given that the buffer now always expects double-precision inputs.

nusbaume (changes requested):
Thanks for bringing in the new history accumulation features (and fixing some bugs) @peverwhee!  I have some questions and change requests, but if you run into problems with any of them then just let me know.  Thanks again!

src/history/cam_hist_file.F90
peverwhee: created https://github.com/ESMCI/history_output/issues/10

src/history/cam_history.F90
peverwhee: created https://github.com/ESCOMP/CAM-SIMA/issues/501

cime_config/testdefs/testlist_cam.xml
peverwhee: good call! done.

cime_config/testdefs/testmods_dirs/cam/outfrq_kessler_mpas_derecho_history/user_nl_cam
peverwhee: ah, yes. good idea. it errors pretty far along in the process in CAM unfortunately. But I added some extra logic to hist_config.py to error during `preview_namelists` for CAM-SIMA

src/history/cam_hist_file.F90
peverwhee: nope! removed.

src/history/cam_hist_file.F90
peverwhee: done

src/history/cam_hist_file.F90
peverwhee: done

src/history/cam_hist_file.F90
```diff
@@ -1556,23 +1555,19 @@ subroutine config_write_field(this, field, split_file_index, restart, &
       frank = size(field_shape)
       if (frank == 1) then
          if (trim(field_precision) == 'REAL32') then
```
nusbaume: Why have this `REAL32` check here at all?  It looks like the `allocate` statements are now exactly the same.

src/history/cam_hist_file.F90
```diff
@@ -1556,23 +1555,19 @@ subroutine config_write_field(this, field, split_file_index, restart, &
...
+            allocate(field_data(end_dims(1) - beg_dims(1) + 1, 1), stat=ierr, errmsg=errmsg)
+            call check_allocate(ierr, subname, 'field_data', file=__FILE__, line=__LINE__-1, errmsg=errmsg)
          else
-            allocate(field_data_r8(end_dims(1) - beg_dims(1) + 1, 1), stat=ierr, errmsg=errmsg)
-            call check_allocate(ierr, subname, 'field_data_r8', file=__FILE__, line=__LINE__-1, errmsg=errmsg)
+            allocate(field_data(end_dims(1) - beg_dims(1) + 1, 1), stat=ierr, errmsg=errmsg)
+            call check_allocate(ierr, subname, 'field_data', file=__FILE__, line=__LINE__-1, errmsg=errmsg)
          end if
       else
          if (trim(field_precision) == 'REAL32') then
```
nusbaume: Same question here, is this if-statement needed anymore?

nusbaume (changes requested):
Thanks for addressing my concerns @peverwhee!  I had two remaining cleanup requests, but after that we should be good to go.  Thanks again!

src/history/cam_hist_file.F90
peverwhee: ah, good point! removed.

src/history/cam_hist_file.F90
peverwhee: removed.

nusbaume (approved):
Everything looks great to me now.  Thanks again @peverwhee!

cime_config/hist_config.py
```diff
@@ -685,6 +696,32 @@ def remove_fields(self, fields, pobj, logger):
...
+        if self.__avg_fields.num_fields() > 0:
+            self.__inst_fields.add_fields(self.__avg_fields.field_names, pobj, [self.__inst_fields], logger)
+            avg_field_list = list(self.__avg_fields.field_names)
+            self.__avg_fields.remove_fields(avg_field_list, pobj, logger)
+        # end if
+
+        if self.__min_fields.num_fields() > 0:
+            self.__inst_fields.add_fields(self.__min_fields.field_names, pobj, [self.__inst_fields], logger)
+            min_field_list = list(self.__min_fields.field_names)
+            self.__min_fields.remove_fields(min_field_list.field_names, pobj, logger)
```
jimmielin: This line appears to be inconsistent with the ones above it, maybe there is an extra `.field_names` attribute here?

src/history/cam_hist_file.F90
```diff
@@ -1515,12 +1513,14 @@ end subroutine config_write_time_dependent_variables
...
-      use hist_buffer,         only: hist_buffer_t
-      use hist_api,            only: hist_buffer_norm_value
+      use hist_api,            only: hist_field_norm_value
       use cam_grid_support,    only: cam_grid_write_dist_array
       use cam_abortutils,      only: check_allocate, endrun
       use hist_field,          only: hist_field_info_t
       use shr_kind_mod,        only: r4 => shr_kind_r4
+      use hist_msg_handler,    only: hist_log_messages
+      use cam_logfile,         only: iulog
+      use cam_abortutils,      only: endrun
```
jimmielin: The `endrun` use statement appears to be repeated in 1518 and 1523.

cime_config/hist_config.py
```diff
@@ -794,6 +831,11 @@ def set_output_frequency(self, ofreq, pobj, logger):
                 ctx = context_string(pobj)
                 logger.debug(f"Setting output_frequency to '{ofreq}'{ctx}")
             # end if
+            # Convert accumulated fields to instantaneous fields if we're set
+            # to output every timestep
+            if self.__output_freq == (1, 'nsteps'):
```
jimmielin: Do we need to check against the other time periods keywords defined in _TIME_PERIODS here:
https://github.com/peverwhee/CAM-SIMA/blob/52df52882c0a1873f3c32f4a623b517999df0da5/cime_config/hist_config.py#L39-L43

```python
# Note, these lists should match the corresponding lists in
#    cam_hist_config_file.F90
_TIME_PERIODS = ['nsteps', 'nstep', 'nseconds', 'nsecond',
                 'nminutes', 'nminute', 'nhours', 'nhour', 'ndays',
                 'nday', 'nmonths', 'nmonth', 'nyears', 'nyear',
                 'steps', 'seconds', 'minutes', 'hours',
                 'days', 'months', 'years']
```

e.g., `1*nstep`, `1*steps`?

jimmielin (changes requested):
Thanks @peverwhee! A few minor fixes for your consideration.

cime_config/hist_config.py
peverwhee: good catch - removed extra `field_names`

src/history/cam_hist_file.F90
peverwhee: redundancy un-redunded

cime_config/hist_config.py
peverwhee: ah - a wonderful catch. added the extra checks. Thank you!

jimmielin (approved):
Thanks @peverwhee!

PR #462, opened 2026-01-13
Bugfix for registry variables with dimensions that are themselves registry variables

peverwhee:
Tag name (required for release branches): 
Originator(s): peverwhee, jimmielin

Description (include the issue title, and the keyword ['closes', 'fixes', 'resolves'] followed by the issue number):
Bugfix for dimensions that are defined in the registry and then used to allocate a different registry variable.

Describe any changes made to build system:
```
M    src/data/generate_registry_data.py
- use the local name for allocating dimensions defined in the registry (but use standard name for metadata)
```
Describe any changes made to the namelist: n/a

List any changes to the defaults for the input datasets (e.g. boundary datasets): n/a

List all files eliminated and why: n/a

List all files added and what they do:
```
A  test/unit/python/sample_files/physics_types_self_ref_dim.F90
A  test/unit/python/sample_files/physics_types_self_ref_dim.meta
A  test/unit/python/sample_files/reg_good_self_ref_dim.xml
- add new test that exercises dimension handling
```
List all existing files that have been modified, and describe the changes: 
(Helpful git command: `git diff --name-status development...<your_branch_name>`)

```
M   test/unit/python/test_registry.py
- add new test that exercises dimension handling
```

If there are new failures (compared to the `test/existing-test-failures.txt` file),
have them OK'd by the gatekeeper, note them here, and add them to the file.
If there are baseline differences, include the test and the reason for the
diff. What is the nature of the change? Roundoff?

derecho/intel/aux_sima: all b4b, except expected NLFAIL for multitape

derecho/gnu/aux_sima: all b4b, except expected FAIL for FADIAB

derecho/nvhpc/aux_sima: run through workflow - b4b

If this changes climate describe any run(s) done to evaluate the new
climate in enough detail that it(they) could be reproduced:

CAM-SIMA date used for the baseline comparison tests if different than latest:

PR #494, opened 2026-04-23
Allow for dimensions set to be used in variables in registry.xml

### LINKED ISSUE #390: Registry should allow for registered variables to also be dimensions of other variables
### What is the feature/what would you like to discuss?

Many registry variables have a `dimensions` attribute, with most of those dimensions being defined by the CCPP-framework itself.  However, it would be beneficial to have the ability to use other registered variables (assuming they are scalar integers with pre-set values) as dimensions when need be, especially for variables that might have specialized or unique dimensions that aren't generally used elsewhere.

Doing this would require `generate_registry_data.py` to do the following:

1.  Check if the dimension names already exist in the registry variable dictionary (if not already one of the official CCPP-framework dimensions)  
2.  If it does exist in the dictionary, check that the variable is an integer type with no dimensions (i.e. a scalar)
3.  Also, check that the variable is either declared a parameter, or has an `initial_value` attribute.
4.  Finally, if all those checks pass, then the script will need to modify the auto-generated code so that the allocation is done with the provided registry variable(s).

Once this work is done, a couple of unit tests could be added to ensure that it works as expected.

### Is there anyone in particular you want to be part of this conversation?

@peverwhee @cacraigucar 

### Will this change (regression test) answers?

No

### Will you be implementing this enhancement yourself?

### LINKED ISSUE #465: CAM-SIMA cannot read physics snapshot variables with non-standard dimensions
### What happened?

Certain physics schemes (e.g. PUMAS) require variables from snapshot files that contain a dimension that is not a member of the standard set of `horziontal_dimension`, `vertical_layer_dimension`, `vertical_interface_dimension`, or `number_of_ccpp_constituents`.  This causes the file read step to fail with an error like so:

```
Cannot read nacon_in from file, nacon_in has unsupported dimension, dust_size_bins_dimension.
```

Given that there will be needed variables from a snapshot with non-standard dimensions, this bug needs to be fixed.

You can replicate this error by doing the following on Derecho:

```
git clone https://github.com/nusbaume/CAM-SIMA.git
cd CAM-SIMA
git checkout pumas_round3
bin/git-fleximod update
```

Then create a new FPHYStest (snapshot) case with the `pumas` test suite, and then build and run the case with the ` cam_ne3pg3_fhistc_ltso_pumas_snapshot_derecho_gnu_*` snapshots found here:

`/glade/campaign/cesm/community/amwg/sima_baselines/cam_sima_test_snapshots`

along with the number of vertical levels set to 58.

The model should then fail with the following error:

` Cannot read nacon_in from file, nacon_in has unsupported dimension, dust_size_bins_dimension.`

2af61da (on pumas_round3 branch)

CISL machine (e.g. cheyenne)

GNU

/glade/derecho/scratch/nusbaume/model_runs/SIMA_pumas_test

Yes

### Extra info

This can (theoretically) be fixed by creating a more generic data-reading interface that doesn't check what dimensions are present, but instead just ensures that the dimension sizes and rank matches the provided buffer variable.

Then the auto-generated initialization code can check if a variable has a non-standard dimension, and if so then use the generic interface instead of the default standard interface.  This should not only fix the bug, but also protect the more precise error messages for all the other variables that do not have an extra dimension.

In the long-term, this generic interface could sit below the more specific interfaces, which would reduce the total amount of code.  However, this is just a "nice-to-have", and not required for fixing this particular bug.

Also pinging @cacraigucar @peverwhee as they helped discover this bug and come up with the potential solution.

jimmielin:
Originator(s): @jimmielin 
  What: claude-opus:4.6[1m]
  How: scoped out the work, wrote tests and proposed fix to horizontal_dimensions unassigned in test

- Fixes #390 
- Fixes #465: Fix self-referential custom dimensions in registry autogen

Background: Variables used in `microp_aero` like `nacon` and `rndst` in `registry.xml` (to be read from snapshot file) use `dust_size_bin_dimension` as a third dimension. That dimension is itself defined as a variable `ndust` (dust_size_bin_dimension) in the same generated module. 

Problem: The autogen code (`generate_registry_data.py`) emits allocate calls using standard names as Fortran identifiers (e.g., allocate(rndst(..., dust_size_bin_dimension))), but only knows how to import dimension variables from external metadata files via use statements. For dimensions defined in the same module, no use statement can be generated (Fortran can't use its own module), so the standard name is undefined at compile time.

This fix generates local integer aliases in the allocate subroutine for any dimension that resolves to a variable in the same module:
```fortran
  integer :: dust_size_bin_dimension  ! local alias
  dust_size_bin_dimension = ndust     ! set before allocate calls
```

  This only triggers when the dimension is not already provided by an external metadata file (i.e., not in `var_module_dict`). The existing `number_of_ccpp_constituents` special case (quoted below) is left untouched:
https://github.com/ESCOMP/CAM-SIMA/blob/a9a626387a0279e96cd5455d303a0da7d3505f0c/src/data/generate_registry_data.py#L1400-L1408

- Fix undefined `horizontal_dimension` variable in tests - Claude pointed out that it was unassigned in the test file. In real builds, `horizontal_dimension` comes from `physics_grid.meta` via a `use` statement, so no alias is generated for it. It should only be fixed in the test sample file.

```
M       src/data/generate_registry_data.py
  - now assign self-referencing dimensions from registry

M       test/unit/python/sample_files/physics_types_parameter.F90
M       test/unit/python/sample_files/physics_types_simple.F90
  - bugfix for unassigned horizontal_dimensions in original test sample file

A       test/unit/python/sample_files/physics_types_self_ref_dim.F90
A       test/unit/python/sample_files/physics_types_self_ref_dim.meta
A       test/unit/python/sample_files/reg_good_self_ref_dim.xml
M       test/unit/python/test_registry.py
  - dust third dimension specified in registry.xml + variable using that dimension in registry.xml (self-referencing) test in test_registry test

M       src/data/registry.xml
  - add dust_size_bin_dimension set to 4
  - add nacon, rndst variables which use new ndust=4 dimension
```

derecho/intel/aux_sima:

derecho/gnu/aux_sima:

### DIFF
--- a/src/data/generate_registry_data.py
+++ b/src/data/generate_registry_data.py
@@ -1405,6 +1405,15 @@ class File:
         #then this section of code will likely need to be modified:
         if ('number_of_ccpp_constituents' in self.__var_dict.known_dimensions):
             outfile.write("use cam_constituents,   only: number_of_ccpp_constituents=>num_constituents", 2)
+
+        # Collect self-referential dimensions (defined in this module)
+        self_ref_dims = []
+        for dim in sorted(self.__var_dict.known_dimensions):
+            if dim not in var_module_dict and dim != 'number_of_ccpp_constituents':
+                var = self.__var_dict.find_variable_by_standard_name(dim)
+                if var and var.local_name != dim:
+                    self_ref_dims.append((dim, var.local_name))
+
         outfile.blank_line()
 
         # Dummy arguments
@@ -1425,6 +1434,8 @@ class File:
         outfile.write(f'logical                     :: {reall_var}', 2)
         subn_str = f'character(len=*), parameter :: subname = "{subname}"'
         outfile.write(subn_str, 2)
+        for dim_std, _ in self_ref_dims:
+            outfile.write(f'integer                     :: {dim_std}', 2)
         outfile.write('', 0)
         outfile.write('! Set optional argument values', 2)
         outfile.write(f'if (present({init_var}_in)) then', 2)
@@ -1437,6 +1448,11 @@ class File:
         outfile.write('else', 2)
         outfile.write(f'{reall_var} = .false.', 3)
         outfile.write('end if', 2)
+        if self_ref_dims:
+            outfile.blank_line()
+            outfile.write('! Set self-referential dimension variables', 2)
+            for dim_std, dim_loc in self_ref_dims:
+                outfile.write(f'{dim_std} = {dim_loc}', 2)
         outfile.write('', 0)
         for var in self.__var_dict.variable_list():
             var.write_allocate_routine(outfile, 2, init_var, reall_var, '', physconst_vars)
--- a/src/data/registry.xml
+++ b/src/data/registry.xml
@@ -1491,6 +1491,18 @@
       <ic_file_input_names>pbuf_VPWP_CLUBB_GW</ic_file_input_names>
     </variable>
 
+    <!-- Aerosol model properties -->
+    <variable local_name="ndust"
+              standard_name="dust_size_bin_dimension"
+              units="count" type="integer">
+      <long_name>number of dust size bins used in aerosol model</long_name>
+      <initial_value>4</initial_value>
+      <!-- this is not runtime configurable, and
+           not even easily compile-time configurable, because the assumption that
+           there are four dust bins is deeply embedded in the microphysics and
+           coupling to the land model (see dstdry1/2/3/4, dstwet1/2/3/4) -->
+    </variable>
+
     <!-- Gravity wave variables - need to be kept in the CAM registry -->
     <variable local_name="kvt"
               standard_name="molecular_kinematic_temperature_conductivity_at_interfaces"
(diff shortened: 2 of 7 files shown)


jimmielin:
I am not sure if this fully resolves #465 but I suspect it does.

OK confirmed it does! I can check nacon and rndst dimensioned (ncol, nlev, dst)

```
 ********** Physics Check Data Results **********

 TIMESTEP:            1
INFLD_REAL8_3D: field = pbuf_NACON, grid = physgrid, timelevel = 1
 Creating new decomp: 486!58!4!!486!58!4!!d6!i1!
 cam_pio_newdecomp: dims =          486          58           4
INFLD_REAL8_3D: field = pbuf_RNDST, grid = physgrid, timelevel = 1

 No differences found (above the threshold) for all variables below:
 Note: If a variable is not in the registry,
       or if a constituent is not registered,
       or the variable was not updated by any scheme,
       it is not checked against the snapshot.
       Verify all expected model state variables are enumerated below:

 Variable Checked                                  # Values   Avg (model)   Avg (snapshot)
 --------                                          --------   ------------   --------------
dust_number_concentration_by_size_bin_for_contact_freezing
                                                     112752    6.00165E+03    6.00165E+03
 dust_radii_by_size_bin                              112752    1.39425E-06    1.39425E-06
```

jimmielin:
It appears that I have independently discovered https://github.com/ESCOMP/CAM-SIMA/pull/462

PR #499, opened 2026-05-14
Fix ic_name_len not accounting for registry constituent IC names

jimmielin:
Originator(s): @jimmielin
  What: claude-opus:4.6[1m]
  How: identified the bug, proposed the fix

* `write_ic_params` in `write_init_files.py` computes `max_loclen` (which sets `ic_name_len` and the `character(len=...)` for `input_var_names`) only from host model variables, not registry constituents. If a constituent's IC name is longer than any host variable's IC name, the generated Fortran array declaration is too short, producing a compile error:
```
Error: Different CHARACTER lengths (12/14) in array constructor at (1)
```
This only happened now because most SDFs pull in host variables with long IC names (e.g., `inverse_exner_function_wrt_surface_pressure`, 43 chars) that mask this bug which only happens with very tiny SDFs.

The existing sample file, incidentally, was already non-buildable because of this issue. It has now been updated.

```
M  src/data/write_init_files.py
  - also check registry_constituents to extend max_loclen with constituent IC name lengths
M  test/unit/python/sample_files/write_init_files/phys_vars_init_check_cnst.F90
  - Update reference file: ic_name_len 12 -> 13, input_var_names padded accordingly
```

derecho/intel/aux_sima:
  SMS_Ln9.ne3pg3_ne3pg3_mg37.FKESSLER.derecho_intel.cam-outfrq_se_cslam_multitape (Overall: NLFAIL) details: pre-existing failure

derecho/gnu/aux_sima:
  SMS_Ln9.ne3pg3_ne3pg3_mg37.FADIAB.derecho_gnu.cam-outfrq_se_cslam (Overall: FAIL) details: pre-existing failure


peverwhee (approved):
thanks for finding and fixing this, @jimmielin !

src/data/write_init_files.py
```diff
@@ -431,6 +431,13 @@ def write_ic_params(outfile, host_vars, ic_names, registry_constituents):
             ic_names[stdname] = [locname]
         # end if
     # end if
```
nusbaume: Not related to your changes, but I think this is supposed to say `for`:


```suggestion
    # end for
```

nusbaume (approved):
Thanks @jimmielin!  I just had one optional request.

Issue #382, opened 2025-04-18
Enable restart history files


peverwhee:
### Issue Type

Infrastructure Update

### Issue Description

CAM-SIMA should have the ability to generate restart history files.

### Will this change answers?

Yes

### Will you be implementing this yourself?

Yes

PR #514, opened 2026-06-23
Restart writing in CAM-SIMA

peverwhee:
Tag name (required for release branches): TBD
Originator(s): peverwhee
AI tools used: None

### PR for initializing and writing restart files (.r. file, rpointer file, and any .rhX. files, as needed)
Addresses #382

Top-level restart module: `src/control/cam_restarts.F90`

🟠**RESTART PHYSICS**🟠
- Variables labeled with `restart=true` in the registry are candidates for the restart file
- New code generates `bld/atm/obj/phys_restart/restart_physics.F90`, which contains "restart_physics_init", "restart_physics_write", and "restart_physics_read" (currently empty)
  - Mods to cime_config python scripts, and a new python script `src/data/write_restart_physics.py` enable this

🟢**RESTART DYNAMICS**🟢
- New restart_dynamics.F90 in both SE and MPAS (`src/dynamics/<dycore>/restart_dynamics.F90`)
  - SE restart dynamics are basically identical to CAM
  - MPAS uses @kuanchihwang 's refactored subdriver to trim the write routine down to one line!

🟣**RESTART HISTORY**🟣
- A static set of variables are written to the restart file (.r.) to indicate the history configuration (this is where we greatly diverge from CAM in terms of variable names!)
- If there are accumulated history fields on a given file (and we haven't JUST written to the file), then a .rhX. file is created.

And finally, a very crude diagram!
Restart writes drawio (1)

```
M   cime_config/buildlib
- Build new phys_restart directory
```

```
A   src/control/cam_restart.F90
- Top-level restart API. Called from cam_timestep_final; calls necessary subroutines to write the restart file (all pieces)

A   src/data/write_restart_physics.py
- Generated code to write variables identified in registry as restart variables - RESTART PHYSICS

A   src/dynamics/mpas/restart_dynamics.F90
A   src/dynamics/se/restart_dynamics.F90
- RESTART DYNAMICS
- note: MPAS init and write are combined so the init phase is a no-op

A   src/history/cam_hist_restart.F90
- RESTART HISTORY
- handles history fields written to overall restart file; the history-specific (rhX) files are handled in cam_hist_file.F90

A   test/unit/python/test_write_restart_physics.py
A   test/unit/python/sample_files/write_restart_physics
A   test/unit/python/sample_files/shared
- Add unit tests for restart_physics.F90 generation (and move shared files with write_init_files testing to shared directory)
```

```
M   .gitmodules
M   src/history/buffers
- updated history external (including new interfaces needed for restarts)

M   cime_config/cam_autogen.py
M   cime_config/cam_build_cache.py
M   cime_config/cam_config.py
M   src/data/registry.xml
M   src/data/registry_v1_0.xsd
M   src/data/generate_registry_data.py
- modifications for physics restarts

M   src/history/cam_hist_file.F90
M   src/history/cam_history.F90
M   src/history/cam_history_support.F90
- modifications for history restarts
- reorganization of write routines in cam_hist_file so code can be shared between history writes and restart writes (and eventually initial conditions file, hopefully).

M   src/control/cam_comp.F90
M   src/cpl/nuopc/atm_comp_nuopc.F90
- Move restart from cam_run4 to timestep_final

M   src/utils/cam_grid_support.F90
- Add 1D integer writing (used for writing sample array)

M   test/existing-test-failures.txt
- Add ERS expected failure
```

derecho/intel/aux_sima:
```
ERS_Ln9.mpasa480_mpasa480.QPC4.derecho_intel.cam-outfrq_analy_ic_cam (Overall: FAIL)
- expected failure until restarts reads are implemented
```

derecho/gnu/aux_sima:
```
ERS_Ln9.mpasa480_mpasa480.QPC4.derecho_gnu.cam-outfrq_analy_ic_cam (Overall: FAIL)
- expected failure until restarts reads are implemented
```

derecho/nvhpc/aux_sima (test is run via Github workflow. Only run the test manually if we need to save new baselines): TBD


cime_config/cam_config.py
```diff
@@ -907,6 +908,19 @@ def generate_cam_src(self, gen_fort_indent):
...
+        #---------------------------------------------------------
+        # Create restart init/write/read routines
+        #---------------------------------------------------------
+        restart_dir = generate_restart_routines(build_cache, self.__bldroot,
+                                                force_ccpp, force_init,
+                                                source_mods_dir, gen_fort_indent,
+                                                capgen_db, ic_names,
+                                                registry_constituents, restart_vars)
+
+        #Add registry path to config object:
```
nusbaume, line 920: Should the comment say "restart" path here?

```suggestion
        #Add physics restart directory path to config object:
```

Also if possible could you update the comment for the `init_dir` config variable addition above as well?

cime_config/cam_autogen.py
```diff
@@ -808,6 +810,54 @@ def generate_init_routines(build_cache, bldroot, force_ccpp, force_init,
...
+            #"restart_physics.py" was modified:
+            do_gen_restart = build_cache.restart_write_mismatch(gen_restart_file)
+    else:
+        #If no directory exists, then one will need
+        # to create new routines:
+        os.mkdir(restart_dir)
+        do_gen_restart = True
+    # End if
+    if do_gen_restart:
+        source_paths = [source_mods_dir, _REG_GEN_DIR]
```
nusbaume, line 846: It doesn't look like `source_paths` is actually used anywhere?

cime_config/cam_autogen.py
```diff
@@ -808,6 +810,54 @@ def generate_init_routines(build_cache, bldroot, force_ccpp, force_init,
 
     return init_dir
 
+###############################################################################
+def generate_restart_routines(build_cache, bldroot, force_ccpp, force_restart,
+                              source_mods_dir, gen_fort_indent, cap_database,
+                              ic_names, registry_constituents, restart_vars):
```
nusbaume, line 816: If you don't need `source_paths` then I think you can simplify this function call to the following:


```suggestion
def generate_restart_routines(build_cache, bldroot, force_ccpp, force_restart,
                              gen_fort_indent, cap_database,
                              registry_constituents, restart_vars):
```

src/control/cam_comp.F90
```diff
@@ -467,51 +467,24 @@ end subroutine cam_run3
    !-----------------------------------------------------------------------
    !
 
-   subroutine cam_run4(rstwr, nlend,                         &
-        yr_spec, mon_spec, day_spec, sec_spec)
+   subroutine cam_run4(rstwr, nlend)
```
nusbaume, line 470: It looks like `cam_run4` isn't doing anything anymore.  Would it be worth just removing it entirely?  We can certainly make an issue to remove this later if that's easier.  Thanks!

src/control/cam_comp.F90
```diff
@@ -521,15 +494,20 @@ subroutine cam_timestep_final(rstwr, nlend, do_ncdata_check, do_history_write)
...
+      use cam_restart,  only: cam_write_restart
       logical, intent(in)  :: rstwr    ! write restart file
       logical, intent(in)  :: nlend    ! this is final timestep
       !Flag for whether a snapshot (ncdata) check should be run or not
       ! - flag is true if this is not the first or last step
       logical, intent(in)  :: do_ncdata_check
+      integer,         intent(in), optional :: yr_spec  ! Simulation year
+      integer,         intent(in), optional :: mon_spec ! Simulation month
+      integer,         intent(in), optional :: day_spec ! Simulation day
+      integer,         intent(in), optional :: sec_spec ! Secs in current simulation day
```
nusbaume, line 506: Total nit-pick, but why is the white-spacing here different from `do_history_write` immediately below here?

Also might be good to just spell out `Seconds`:


```suggestion
      integer,         intent(in), optional :: yr_spec  ! Simulation year
      integer,         intent(in), optional :: mon_spec ! Simulation month
      integer,         intent(in), optional :: day_spec ! Simulation day
      integer,         intent(in), optional :: sec_spec ! Seconds in current simulation day
```

src/control/cam_comp.F90
```diff
@@ -540,7 +518,20 @@ subroutine cam_timestep_final(rstwr, nlend, do_ncdata_check, do_history_write)
...
+      !
+      if (rstwr) then
+         call t_startf('cam_write_restart')
+         if (present(yr_spec) .and. present(mon_spec) .and.                   &
+              present(day_spec).and.present(sec_spec)) then
+            call cam_write_restart(dyn_out, yr_spec=yr_spec, &
+                 mon_spec=mon_spec, day_spec=day_spec, sec_spec= sec_spec)
+         else
+            call cam_write_restart(dyn_out)
+         end if
```
nusbaume, line 532: Given that it looks like these fields are `optional` in `cam_write_restart` as well, I believe we can just pass them directly without the need to check them (even if they aren't present here):




```suggestion
         call cam_write_restart(dyn_out, yr_spec=yr_spec, &
                 mon_spec=mon_spec, day_spec=day_spec, sec_spec= sec_spec)
```

src/control/cam_restart.F90
```diff
@@ -0,0 +1,149 @@
...
+    type(io_desc_t), pointer     :: iodesc
+    type(cam_grid_header_info_t) :: info
+
+    ! Set template for primary restart filename based on instance suffix
+    ! (%c = caseid, $y = year, $m = month, $d = day, $s = seconds in day, %t = number)
+    rfilename_spec = '%c.cam' // trim(inst_suffix) //'.r.%y-%m-%d-%s.nc'
+    fname = interpret_filename_spec(rfilename_spec, yr_spec=yr_spec, &
+            mon_spec=mon_spec, day_spec=day_spec, sec_spec= sec_spec)
+
+    call cam_pio_createfile(fh, trim(fname), 0)
```
nusbaume, line 53: It would be worth double-checking me, but I _think_ passing in a zero here as the third argument will do nothing, as an `ior` operation will always just return whatever the other integer is when compared to `0`.   If that's true then I believe this can just be:


```suggestion
    call cam_pio_createfile(fh, trim(fname))
```

src/control/cam_restart.F90
```diff
@@ -0,0 +1,149 @@
...
+
+    ! TODO: initialize ionosphere restart
+
+    ! Initialize physics grid variable
+    grid_id = cam_grid_id('physgrid')
+    call cam_grid_write_attr(fh, grid_id, info)
+
+    call restart_physics_init(fh, errmsg, errflg)
+    if (errflg /= 0) then
+       call endrun(errmsg)
```
nusbaume, line 66: Since we are doing file I/O here should this be a "safe" endrun?


```suggestion
       call safe_endrun(errmsg)
```

src/control/cam_restart.F90
```diff
@@ -0,0 +1,149 @@
...
+
+    ! TODO: write ionosphere restart
+
+    ! Write physics grid info
+    call cam_grid_write_var(fh, phys_decomp)
+
+    ! Write physics restart variables
+    call restart_physics_write(fh, grid_id, errmsg, errflg)
+    if (errflg /= 0) then
+       call endrun(errmsg)
```
nusbaume, line 89: Use `safe_endrun` here?


```suggestion
       call safe_endrun(errmsg)
```

src/control/cam_restart.F90
```diff
@@ -0,0 +1,149 @@
...
+    end if
+
+    call history_restart_write(fh)
```
nusbaume, line 92: Total nit-pick, but it looks like the dynamics, physics, and history restart writing subroutine names all have three different orderings.  Can we unify them to a single order to help with future grepping?  For example:

`write_restart_dynamics`
`write_restart_physics`
`write_restart_history`

instead of

`write_restart_dynamics`
`restart_physics_write`
`history_restart_write`

Of course if you did the different naming conventions on purpose then feel free to ignore this request!

src/control/cam_restart.F90
```diff
@@ -0,0 +1,149 @@
...
+
+    ! Close the primary restart file
+    call pio_closefile(fh)
```
nusbaume, line 95: I would use `cam_pio_closefile` (from `cam_pio_utils`) here so that the file is properly registered as closed:


```suggestion
    call cam_pio_closefile(fh)
```

src/control/cam_restart.F90
```diff
@@ -0,0 +1,149 @@
...
+
+     if (masterproc) then
+        rest_pfile = interpret_filename_spec('rpointer.cam'//trim(inst_suffix)//'.'//'%y-%m-%d-%s',&
+             yr_spec=yr_spec, mon_spec=mon_spec, day_spec=day_spec, sec_spec= sec_spec )
+        call cam_open_file(rest_pfile, unitn, 'f', status="unknown")
+        rewind unitn
+        write(unitn, '(a)', iostat=ierr) trim(restart_file)
+        if (ierr /= 0) then
+           call endrun(sub//': ERROR: writing rpointer file')
+        end if
```
nusbaume, line 136: I would include `iomsg` here as well:

```suggestion
        write(unitn, '(a)', iostat=ierr, iomsg=errmsg) trim(restart_file)
        if (ierr /= 0) then
           call endrun(sub//': ERROR: writing rpointer file: '//errmsg)
        end if
```

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -1183,10 +1183,10 @@ subroutine ModelAdvance(gcomp, rc)
        call t_stopf  ('CAM_run3')
 
        call t_startf ('CAM_run4')
-       call cam_run4( rstwr, nlend, &
-            yr_spec=yr_sync, mon_spec=mon_sync, day_spec=day_sync, sec_spec=tod_sync)
+       call cam_run4(rstwr, nlend)
        call t_stopf  ('CAM_run4')
-       call cam_timestep_final(rstwr, nlend, do_ncdata_check=do_ncdata_check)
+       call cam_timestep_final(rstwr, nlend, do_ncdata_check=do_ncdata_check, &
```
nusbaume, line 1188: I realize this isn't the responsibility of this particular PR, but should we wrap `cam_timestep_final` with `t_startf` and `t_stopf` calls?

src/data/generate_registry_data.py
```diff
@@ -571,6 +579,10 @@ def __init__(self, var_node, known_types, vdict, dycore, logger):
         self.__def_dims_str = ""
         for attrib in var_node:
             if attrib.tag == 'dimensions':
+                if not attrib.text:
+                    emsg = f"Variable '{local_name}' cannot be a restart variable without any dimensions."
+                    raise CCPPError(emsg)
```
nusbaume, line 584: Given the eventual "split" with the CCPP-framework code, should we choose a different exception to raise here?  For example:


```suggestion
                    raise ValueError(emsg)
```

src/data/registry_v1_0.xsd
```diff
@@ -110,10 +110,14 @@
...
   <xs:attribute name="constituent"             type="xs:boolean"
                 default="false"/>
+  <xs:attribute name="restart"             type="xs:boolean"
+                default="false"/>
   <xs:attribute name="advected"                type="xs:boolean"
                 default="false"/>
   <xs:attribute name="phys_timestep_init_zero" type="xs:boolean"
                 default="false"/>
+  <xs:attribute name="diagnostic_name" type="fortran_id_type"
+                default="none"/>
```
nusbaume, line 120: Nit-pick, but should we move this to the section above where the other `fortran_id_type` attributes are?  Alternatively we can make a new `optional attributes` section for attributes which aren't boolean but are optional.

src/data/registry.xml
```diff
@@ -139,14 +139,16 @@
     <variable local_name="frontgf"
               standard_name="frontogenesis_function"
               units="K2 m-2 s-1" type="real" kind="kind_phys"
-              allocatable="pointer">
+              allocatable="pointer" restart="true"
+	      diagnostic_name="FRONTGF">
```
nusbaume, line 143: Definite nit-pick, but is there a reason `diagnostic_name` and/or `allocatable` is always indented differently then all the other attributes here and below?  Of course if this is just a Github rendering issue then feel free to ignore this question.  Thanks!

src/data/registry.xml
```diff
@@ -609,7 +616,7 @@
     </variable>
     <variable local_name="cldliq"
               standard_name="cloud_liquid_water_mixing_ratio_wrt_moist_air_and_condensed_water"
-              units="kg kg-1" type="real" constituent="true">
+              units="kg kg-1" type="real" constituent="true" restart="true" diagnostic_name="cldliq">
```
nusbaume, line 619: Should the `diagnostic_name` be `CLDLIQ` here, or does it not matter?

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,600 @@
...
+                          'ccpp_constituent_properties',
+                          'ccpp_constituent_minimum_values',
+                          'ccpp_error_message',
+                          'ccpp_error_code',
+                          'log_output_unit', 'do_log_output',
+                          'mpi_communicator', 'mpi_root', 'mpi_rank',
+                          'number_of_mpi_tasks'}
+# Variable input types
+_INPUT_TYPES = set(['in', 'inout'])
+_OUTPUT_TYPES = set(['out', 'inout'])
```
nusbaume, line 31: I believe it is more performant to create the set directly:


```suggestion
_INPUT_TYPES = {'in', 'inout'}
_OUTPUT_TYPES = {'out', 'inout'}
```

Can we make the same change to these variables in `write_init_files.py` as well?  Thanks!

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,600 @@
...
+
+# Increase allowed line lengths needed to fit extra-long CCPP standard names:
+_LINE_FILL_LEN = 150
+_MAX_LINE_LEN = 200
```
nusbaume, line 35: Actually, does it make sense to just grab all of these fields from `write_init_files.py`, or have that code grab them from here?  That way there is a single location to modify all of these variables, in case we ever need to in the future.

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,600 @@
...
+        outfile.write("public :: restart_physics_read", 1)
+
+        all_req_vars = in_vars + out_vars
+        # Grab the required variables that are also restart variables
+        host_dict = cap_database.host_model_dict()
+        required_restart_vars, constituent_dimmed_vars, used_vars = gather_required_restart_variables(all_req_vars, restart_vars, host_dict)
+
+        outfile.blank_line()
+        outfile.comment("Private module data", 0)
+        for _, value in required_restart_vars.items():
```
nusbaume, line 105: Here and below, would it make sense to use `values` instead of `items` when looping over the dictionaries (given that we don't need the keys)?


```suggestion
        for value in required_restart_vars.values():
```

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,600 @@
...
+                 ["ccpp_constituent_prop_mod", ["ccpp_constituent_prop_ptr_t"]]]
+
+    # Gather up dimension imports
+    dim_use_stmts = [['vert_coord', ['pver']], ['physics_grid', ['columns_on_task']]]
+    for key in sorted(dim_use_stmt_dict):
+        imports = []
+        for var_import in sorted(dim_use_stmt_dict[key]):
+            imports.append(var_import)
+        # end for
+        dim_use_stmts.append([key, imports])
```
nusbaume, line 196: I believe this can be simplified like so:


```suggestion
        dim_use_stmts.append([key, sorted(dim_use_stmt_dict[key])])
```

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,600 @@
...
+    # end for
+
+    # Output required, registered Fortran module use statements:
+    write_use_statements(outfile, static_use_stmts, 2)
+    write_use_statements(outfile, dim_use_stmts, 2)
+
+    outfile.write("type(file_desc_t), intent(inout) :: file", 2)
+    outfile.write("character(len=512),intent(out)   :: errmsg", 2)
```
nusbaume, line 204: Should this be `len=*`?

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,600 @@
...
+            # end if
+            hdimids.append(dimensions_dict[dimname]['index'])
+        # end for
+        outfile.blank_line()
+        outfile.comment("Define required restart variables on the restart file", 2)
+        desc_name = f"{value['diag_name'].lower()}_desc"
+        dimids_array = ", ".join([f"dimids({i})" for i in hdimids])
+        outfile.write(f"call cam_pio_def_var(file, '{value['diag_name']}', pio_double, (/{dimids_array}/), {desc_name}, existOK=.false.)", 2)
+        outfile.write("if (errflg /= 0) then", 2)
+        outfile.write(f"write(errmsg,*) 'restart_physics_init: error defining variable {value['diag_name']}'", 3)
```
nusbaume, line 259: I would add quotes around the bad variable name here:


```suggestion
        outfile.write(f'''write(errmsg,*) 'restart_physics_init: error defining variable "{value['diag_name']}" ' ''', 3)
```

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,600 @@
...
+            outfile.write("end do", 2)
+            outfile.blank_line()
+        # end for
+    # end if
+
+    # Handle non-advected constituent variables
+    outfile.blank_line()
+    outfile.comment("Handling for non-advected constituent vars (advected constituents handled by dynamics restart)", 2)
+    outfile.comment("Allocate cnst_desc to total size of constituents array; some will be unused", 2)
+    outfile.write("allocate(cnst_desc(size(const_props)))", 2)
```
nusbaume: Instead of using the size of all constituents, I believe we can determine the total number of non-advected constituents before allocating.  For example, if we use `num_constituents ` and `num_advected` from `cam_constituents.F90`, then this could be allocated to `num_constituents - num_advected`.  Additionally we could check if that difference is zero, and if so then skip the allocation and `pio_def_var` loop below entirely (as no non-advected constituents exist).

If you do apply this fix, then we'll also need to add a check to the `restart_physics_write` that only attempts to write these fields if `cnst_desc` has been allocated (i.e. `num_constituents - num_advected > 0`).

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,600 @@
...
+        outfile.comment("Private module data", 0)
+        for _, value in required_restart_vars.items():
+            outfile.write(f"type(var_desc_t) :: {value['diag_name'].lower()}_desc", 1)
+        # end for
+
+        for _, value in constituent_dimmed_vars.items():
+            outfile.write(f"type(var_desc_t), allocatable :: {value['diag_name'].lower()}_desc(:)", 1)
+        # end for
+
+        outfile.write("type(var_desc_t), allocatable :: cnst_desc(:)", 1)
```
nusbaume: Sorry for not seeing this earlier, but given that `restart_physics_init` could potentially be called every timestep, we need to have checks that only allocate these module variables if they haven't already been allocated (e.g. only if `allocated()` returns `.false.`).

cime_config/testdefs/testlist_cam.xml
```diff
@@ -252,6 +252,16 @@
...
+  <test compset="QPC4" grid="mpasa480_mpasa480" name="ERC_Ln9" testmods="cam/outfrq_analy_ic_cam4">
+    <machines>
+      <machine name="derecho" compiler="intel" category="aux_sima"/>
+      <machine name="derecho" compiler="gnu" category="aux_sima"/>
+    </machines>
+    <options>
+      <option name="wallclock">00:20:00</option>
+      <option name="comment">Aquaplanet CAM4 restart test with MPAS dycore at 480 km resolution</option>
+    </options>
+  </test>
```
nusbaume, line 264: I think we'll want to include SE dycore tests as well, along with tests that enable the compiler debug settings (i.e. `ERC_D_Ln9`).  A good testing strategy you brought up in the chat is:

> I think we can do maybe one debug, one non-debug per set per dycore, but flip-flop them? Like DEBUG/SE/GNU and NONDEBUG/SE/INTEL, vs NONDEBUG/MPAS/GNU and DEBUG/MPAS/INTEL?

So can you make those changes here?  Thanks!

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+                        # end if
+                        outfile.write(f"call cam_pio_def_dim(file, '{dim_loc_name}', {dimsize}, dimids({dim_index}), existOK=.true.)", 2)
+                        dimensions_dict[dimname]['index'] = dim_index
+                        dim_index = dim_index + 1
+                    # end if
+                    hdimids.append(dimensions_dict[dimname]['index'])
+                # end if (ignore number of constituents dimensions)
+            # end for
+            outfile.comment(f"Handling for constituent-dimensioned variable '{value['diag_name']}'", 2)
+            outfile.write(f"allocate({value['diag_name'].lower()}_desc(size(const_props)), stat=errflg, errmsg=errmsg)", 2)
```
nusbaume, line 298: Does this `allocate` statement need to be guarded with an `allocated()` check?

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+            outfile.write(f"call cam_pio_def_var(file, '{value['diag_name']}_'//trim(const_diag_name), pio_double, (/{dimids_array}/), {desc_name}(constituent_idx), existOK=.false.)", 3)
+            outfile.write("end do", 2)
+            outfile.blank_line()
+        # end for
+    # end if
+
+    # Handle non-advected constituent variables
+    outfile.blank_line()
+    outfile.comment("Handling for non-advected constituent vars (advected constituents handled by dynamics restart)", 2)
+    outfile.comment("Allocate cnst_desc to total size of constituents array; some will be unused", 2)
```
nusbaume, line 316: Instead of using the size of all constituents, I believe we can determine the total number of non-advected constituents before allocating.  For example, if we use `num_constituents ` and `num_advected` from `cam_constituents.F90`, then this could be allocated to `num_constituents - num_advected`.  Additionally we could check if that difference is zero, and if so then skip the allocation and `pio_def_var` loop below entirely (as no non-advected constituents exist).

If you do apply this fix, then we'll also need to add a check to the `restart_physics_write` that only attempts to write these fields if `cnst_desc` has been allocated (i.e. `num_constituents - num_advected > 0`).

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+    # end if
+    outfile.comment("Write required restart variables to the restart file", 2)
+    for key, value in required_vars.items():
+        desc_name = f"{value['diag_name'].lower()}_desc"
+        if len(value['dims']) == 1 and 'horizontal_dimension' in value['dims'][0]:
+            outfile.comment("Handle horizontal-only field", 2)
+            outfile.write("field_shape(1) = num_global_phys_cols", 2)
+            outfile.write(f"call cam_grid_write_dist_array(file, grid_decomp, (/dims(1)/), (/field_shape(1)/), {key}, {desc_name})", 2)
+        elif len(value["dims"]) > 1:
+            dimstr = '(/'
```
nusbaume: Can we use the more modern array constructor syntax here and below?


```suggestion
            dimstr = '['
```

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+            for index, dim in enumerate(value["dims"]):
+                dimstr = f"{dimstr}dims({dimensions_dict[dim.split(':')[1]]['index']}),"
+                if 'horizontal_dimension' in dim:
+                    outfile.write("field_shape(1) = num_global_phys_cols", 2)
+                else:
+                    outfile.write(f"field_shape({index + 1}) = size({key}, {index + 1})", 2)
+                # end if
+            # end if
+            outfile.write(f"call cam_grid_write_dist_array(file, grid_decomp, {dimstr[:-1]}/), field_shape, {key}, {desc_name})", 2)
```
nusbaume: Just noting that if you accept my previous request then you'll need to modify this as well:


```suggestion
            outfile.write(f"call cam_grid_write_dist_array(file, grid_decomp, {dimstr[:-1]}], field_shape, {key}, {desc_name})", 2)
```

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+    for dim in dimensions_dict:
+        outfile.write(f"dims({dimensions_dict[dim]['index']}) = {dimensions_dict[dim]['import']}", 2)
+    # end if
+    outfile.comment("Write required restart variables to the restart file", 2)
+    for key, value in required_vars.items():
+        desc_name = f"{value['diag_name'].lower()}_desc"
+        if len(value['dims']) == 1 and 'horizontal_dimension' in value['dims'][0]:
+            outfile.comment("Handle horizontal-only field", 2)
+            outfile.write("field_shape(1) = num_global_phys_cols", 2)
+            outfile.write(f"call cam_grid_write_dist_array(file, grid_decomp, (/dims(1)/), (/field_shape(1)/), {key}, {desc_name})", 2)
```
nusbaume, line 388: Here and below, can we use the newer array-constructor syntax when creating arrays?  For example:


```suggestion
            outfile.write(f"call cam_grid_write_dist_array(file, grid_decomp, [dims(1)], [field_shape(1)], {key}, {desc_name})", 2)
```

nusbaume: Also, why set `field_shape` at all here?  Couldn't you just use `num_global_phys_cols` directly?

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+    outfile.blank_line()
+
+    # Handle constituent-dimensioned variables
+    if constituent_dimmed_vars:
+        for key, value in constituent_dimmed_vars.items():
+            outfile.comment(f"Handling for constituent-dimensioned variable '{value['diag_name']}'", 2)
+            outfile.write("do constituent_idx = 1, size(const_props)", 2)
+            desc_name = f"{value['diag_name'].lower()}_desc"
+            outfile.write("field_shape(1) = num_global_phys_cols", 3)
+            outfile.write(f"call cam_grid_write_dist_array(file, grid_decomp, (/dims(1)/), (/field_shape(1)/), {key}(:,constituent_idx), {desc_name}(constituent_idx))", 3)
```
nusbaume, line 413: Can't you just use `num_global_phys_cols` directly here?  Also a reminder to update the array syntax:


```suggestion
            outfile.write(f"call cam_grid_write_dist_array(file, grid_decomp, [dims(1)], [num_global_phys_cols], {key}(:,constituent_idx), {desc_name}(constituent_idx))", 3)
```

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+    """
+    outfile.write("subroutine restart_physics_read()", 1)
+    outfile.write("end subroutine restart_physics_read", 1)
+
+#################
+#HELPER FUNCTIONS
+#################
+
+##############################################################################
+def _find_and_add_host_variable(stdname, host_dict, var_dict):
```
nusbaume, line 449: It seems like this function and `gather_ccpp_req_vars` share a lot of code with the same functions in `write_init_files.py`.  I wonder if there is a way to combine them (maybe with some logical flags to combine the slightly different behaviors) so that we can then have a single set of functions for both routines?

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+        mvlist = ', '.join(sorted(missing_vars))
+        retmsg = f"Error: Missing required host variables: {mvlist}"
+    # end if
+    # Return the required variables as a list
+    return list(in_vars.values()), list(out_vars.values()), retmsg
+
+##############################################################################
+def gather_required_restart_variables(all_req_vars, restart_vars, host_dict):
+    """
+    Return lists of the required restart non-constituent variables, and the required
```
nusbaume, line 545: Should this say return dictionaries?  It might also help to describe what the dictionaries contain in a little more detail here.

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+
+    required_restart_vars = {}
+    required_constituent_dimensioned_vars = {}
+    used_vars = set()
+
+    for required_var in all_req_vars:
+        stdname = required_var.get_prop_value('standard_name')
+        local_name = required_var.call_string(host_dict)
+        used_var = required_var.var.get_prop_value('local_name')
+        if stdname in restart_vars:
```
nusbaume, line 557: Should these two calls be moved under the `if stdname in restart_vars:` check?


```suggestion
        if stdname in restart_vars:
            local_name = required_var.call_string(host_dict)
            used_var = required_var.var.get_prop_value('local_name')
```

test/unit/python/sample_files/shared/ddt2_reg.xml
```diff
@@ -4,12 +4,14 @@
   <file name="physics_types_ddt2" type="module">
```
nusbaume, line 4: Definite nit-pick, but can we change the new `shared` directory name to `shared_files`?  That way it matches all of the other subdirectories that end in `_files`.

test/unit/python/sample_files/shared/simple_host.F90
```diff
@@ -0,0 +1,96 @@
...
+    end do
+
+    call simple_ccpp_physics_timestep_final('simple_suite', errmsg, errflg)
+
+    call simple_ccpp_physics_finalize('simple_suite', errmsg, errflg)
+    if (errflg /= 0) then
+      write(6, *) trim(errmsg)
+      write(6,'(a)') 'An error occurred in ccpp_timestep_final, Exiting...'
+      stop
+    end if
```
nusbaume, line 81: This might not matter as this is a test scheme, but should this be directly under the `timestep_final` call, with then a different `errflg` check for the `finalize` call?

test/unit/python/sample_files/write_restart_physics/restart_physics_const_dim.F90
```diff
@@ -0,0 +1,173 @@
...
+      call cam_pio_def_dim(file, 'ncol', num_global_phys_cols, dimids(1), existOK=.true.)
+      call cam_pio_def_dim(file, 'lev', pver, dimids(2), existOK=.true.)
+      const_props => cam_model_const_properties()
+
+      ! Handling for constituent-dimensioned variable 'cool_cat_for_each_const'
+      allocate(cool_cat_for_each_const_desc(size(const_props)), stat=errflg, errmsg=errmsg)
+      if (errflg /= 0) then
+         return
+      end if
+      do constituent_idx = 1, size(const_props)
```
nusbaume, line 72: If we use `num_constituents` from `cam_constituents.F90` in this generated routine then I believe we can replace all of the `size(const_props)` here and below:


```suggestion
      do constituent_idx = 1, num_constituents
```

test/unit/python/sample_files/write_restart_physics/restart_physics_const_dim.F90
```diff
@@ -0,0 +1,173 @@
...
+         ! Grab constituent diagnostic name:
+         call const_props(constituent_idx)%diagnostic_name(const_diag_name)
+         call cam_pio_def_var(file, 'cool_default_cat_for_each_const_'//trim(const_diag_name), pio_double, (/dimids(1)/), &
+             cool_default_cat_for_each_const_desc(constituent_idx), existOK=.false.)
+      end do
+
+
+      ! Handling for non-advected constituent vars (advected constituents handled by dynamics restart)
+      ! Allocate cnst_desc to total size of constituents array; some will be unused
+      allocate(cnst_desc(size(const_props)))
```
nusbaume, line 94: Should we check the allocate status of the `cnst_desc` allocation, and return if it is non-zero?

test/unit/python/sample_files/write_restart_physics/restart_physics_constituents.F90
```diff
@@ -0,0 +1,117 @@
...
+      ! Define required restart variables on the restart file
+      ! Define potentially new dimension 'horizontal_dimension'
+      call cam_pio_def_dim(file, 'ncol', num_global_phys_cols, dimids(1), existOK=.true.)
+      ! Define potentially new dimension 'vertical_layer_dimension'
+      call cam_pio_def_dim(file, 'lev', pver, dimids(2), existOK=.true.)
+      call cam_pio_def_var(file, 'theta', pio_double, (/dimids(1), dimids(2)/), theta_desc, existOK=.false.)
+      if (errflg /= 0) then
+         write(errmsg,*) 'restart_physics_init: error defining variable theta'
+         return
+      end if
```
nusbaume, line 68: I don't think we need the `errflg` check after the `cam_pio_def_var` calls, as `cam_pio_def_var` will simply error on its own if it encounters a problem and `existOK = .false.`.

test/unit/python/sample_files/write_restart_physics/restart_physics_no_required.F90
```diff
@@ -0,0 +1,127 @@
...
+
+      type(file_desc_t), intent(inout) :: file
+      integer,            intent(in)   :: grid_id
+      character(len=512),intent(out)   :: errmsg
+      integer,           intent(out)   :: errflg
+
+      ! Local variables
+      integer                          :: dims(2)
+      integer                          :: grid_decomp
+      integer                          :: grid_dims(2)
```
nusbaume, line 96: Is `grid_dims` actually needed?  It doesn't appear to ever be used in the generated Fortran code.

test/unit/python/test_write_restart_physics.py
```diff
@@ -0,0 +1,522 @@
...
+
+if not os.path.exists(__REGISTRY_DIR):
+    EMSG = "Cannot find registry directory where 'write_restart_physics.py' should be located."
+    raise ImportError(EMSG)
+
+if not os.path.exists(_REG_SAMPLES_DIR):
+    raise ImportError("Cannot find sample files directory")
+
+if not os.path.exists(_SHARED_DIR):
+    raise ImportError("Cannot find 'write_restart_physics' sample files directory")
```
nusbaume, line 49: Update the error message here?


```suggestion
    raise ImportError("Cannot find 'shared' sample files directory")
```

test/unit/python/test_write_restart_physics.py
```diff
@@ -0,0 +1,522 @@
...
+
+#Add registry directory to python path to import
+#registry and 'phys_restart' code generators:
+sys.path.append(__REGISTRY_DIR)
+
+# pylint: disable=wrong-import-position
+from ccpp_capgen import capgen
+from framework_env import CCPPFrameworkEnv
+from generate_registry_data import gen_registry
+import write_restart_physics as write_restart
```
nusbaume, line 66: Just curious why this line isn't `from write_restart_physics import write_restart_physics` like all of the other import statements?

src/data/write_restart_physics.py
```diff
@@ -0,0 +1,602 @@
...
+                    # do nothing with missing variables
+                # end if
+            # end if (only output variables)
+        # end for (loop over call list)
+    # end for (loop over phases)
+
+    if missing_vars:
+        mvlist = ', '.join(sorted(missing_vars))
+        retmsg = f"Error: Missing required host variables: {mvlist}"
+    # end if
```
nusbaume, line 538: I don't know how feasible it is, but it might be good to have a unit test that ensures this error message is returned correctly if there is a missing required host variable.

nusbaume (changes requested):
Thanks for bringing in this major capability @peverwhee!  Due to shifting priorities I'll need to take a break on reviewing this PR, so I wanted to submit what I had now, which should cover the full physics restart writing capability, but not the dynamics or history restarts.  Of course if you have any questions or concerns with what I've commented on so far just let me know.  Thanks again!

src/history/cam_hist_file.F90
```diff
@@ -1380,42 +1347,834 @@ subroutine config_define_file(this, restart, logname, host, model_doi_url)
...
+         if (this%field_list(field_index)%accumulate_type() == 'var') then
+            write_var_buffer = .true.
+            var_buffer_idx = var_buffer_idx + 1
+         else
+            write_var_buffer = .false.
+         end if
+         call this%write_restart_fields(this%restart_file, this%field_list(field_index), field_index, write_var_buffer, &
+            var_buffer_idx)
+         ! Clear the buffers
+         call this%field_list(field_index)%clear_buffers()
```
jimmielin, line 2138: Quick question! I wonder if we should clear the buffers here. Aren't restarts instantaneous? After writing a restart file wouldn't this clear the buffers for other accumulated fields as well or are history tapes tracked separately?

src/history/cam_hist_restart.F90
```diff
@@ -0,0 +1,442 @@
...
+          long_name(1:num_fields(idx), idx) = hist_configs(idx)%get_long_names()
+          ! Get units
+          units(1:num_fields(idx), idx) = hist_configs(idx)%get_units()
+          ! Get fill value flag and fill value
+          fill_flag(1:num_fields(idx), idx) = hist_configs(idx)%get_fill_flags()
+          fill_value(1:num_fields(idx), idx) = hist_configs(idx)%get_fill_values()
+          ! Get field dimension indices
+          ndims(1:num_fields(idx)) = hist_configs(idx)%get_num_dimensions()
+          do jdx = 1, num_fields(idx)
+             dimensions(1:ndims(jdx), 1:num_fields(idx), idx) = hist_configs(idx)%get_dimension_indices()
```
jimmielin, line 185: It looks like `dimids`'s first dimension get hardcoded to `4` in `config_get_dimension_indices()`. In this case if `ndims(jdx) /= 4` I think this might write out of bounds in memory

src/dynamics/se/restart_dynamics.F90
```diff
@@ -0,0 +1,563 @@
...
+      call PIO_Setframe(file, dp_fvm_desc, t_idx)
+      call PIO_Write_Darray(file, dp_fvm_desc, iodesc3d_fvm, buf3d, ierr)
+
+      do m = 1, size(const_props)
+         do ie = 1, nelemd
+            do k = 1, nlev
+               ii = 1
+               do j = 1, nc
+                  do i = 1, nc
+                     buf3d(ii,k,ie) = fvm(ie)%c(i,j,k,m)
```
jimmielin, line 253: I think here and throughout might need to use `qsize` or `ntrac`? `const_props` might have a lot of non-advected constituents but the dycore only sees advected tracers

jimmielin (commented):
Not an official reviewer but tried to run this out of curiosity and had a few questions!

PR #517, opened 2026-07-02
Allow excluding fields from ncdata_check

jimmielin:
Originator(s): @jimmielin 
  What: claude-fable:5
  How: Discussed the implementation, wrote most of the logic and tests.

- Allows excluding certain fields based on a `.gitignore`-like glob pattern from `physics_check_data`.
- This feature is expected to be used so `ncdata_check_err=.true.` can be turned on in tests that read multi-stage snapshots where some variables are explicitly intended to be ignored as they would not match.

Describe any changes made to the namelist: host-side `ncdata_check_exclude` option - default empty

```
M       cime_config/namelist_definition_cam.xml
  - add ncdata_check_exclude

M       src/core_utils/string_core_utils.F90
  - pure core_glob_match and core_glob_list_excluded

M       src/physics/utils/phys_comp.F90
  - add &physics_nl ncdata_check_exclude

M       src/physics/utils/physics_data.F90
  - check_field_2d/3d/4d skip rows whose report label matches the exclusion list
  - write out the excluded list always so it is not a silent drop.

M       test/unit/fortran/src/core_utils/test_string_core_utils.pf
  - comprehensive tests for the added glob matching string utils.
```

derecho/intel/aux_sima:

derecho/gnu/aux_sima:

### DIFF
--- a/cime_config/namelist_definition_cam.xml
+++ b/cime_config/namelist_definition_cam.xml
@@ -206,6 +206,39 @@
       <value>.false.</value>
     </values>
   </entry>
+  <entry id="ncdata_check_exclude">
+    <type>char*256(200)</type>
+    <category>initial_conditions</category>
+    <group>physics_nl</group>
+    <desc>
+      Ordered list of glob patterns ('*' matches any run of characters)
+      excluding variables from the `{{ hilight }}ncdata_check{{ closehilight }}`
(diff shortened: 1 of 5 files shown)


nusbaume (approved):
Applied what would have been my change requests in the latest commit!

peverwhee (approved):
committed the one small comment change i had.

PR #518, opened 2026-07-06
sima0_16_001: Make constituent names case-insensitive in IC file read path, consistent with the CCPP framework; update framework tag

jimmielin:
Originator(s): @jimmielin 
  What: see below
  How: claude-fable:5 identified and drafted the original case-insensitive fix; claude-opus:4.8 refined the fix

- Case-insensitive constituent `ic_file_input_name` matching. Standard names are case-insensitive and capgen lowercases scheme-side names (e.g. `co2`), but the IC read code compared them case-sensitively so a registry constituent with authored case (`CO2`, `O3`) or an uppercase snapshot field read nothing (could see this failure in `atm.log...` it falls back to the argument name). All constituent-name comparisons are now case-insensitive.
- All-case cnst_/pbuf_ fallback. For constituents not enumerated in the registry, the read/check fallback now tries `{std_name, cnst_, pbuf_} × {as-is, UPPER, lower}` so a lowercase runtime name (`co2`) still matches an uppercase snapshot field (`CO2`, `cnst_CO2`). netCDF variable names are case-sensitive. CAM has uppercase for most chemistry and physics fields and lowercase for aerosols (e.g., `num_a1`)
- Suppress local-name rows for non-registry constituents. `gather_ccpp_req_vars` no longer adds constituents to the required host-variable list, which previously gave them a `phys_var_stdnames` row keyed to the scheme's local Fortran name (e.g. `co2mmr`) that preempted the cnst_/pbuf_ fallback and left the constituent at its default.

```
M  src/data/cam_var_init_marks.inc
  - Case-insensitive mark_as_*/is_* standard-name lookups.

M  src/data/write_init_files.py
  - Generated phys_read_data/phys_check_data now match constituent standard
    names case-insensitively and try all-case cnst_/pbuf_ candidates in the
    not-in-registry fallback.
  - gather_ccpp_req_vars no longer adds constituents to in_vars so the fallback
    can actually be reached for meta-declared constituents.

M  src/physics/utils/physics_data.F90
  - Case-insensitive standard-name matching in find_input_name_idx and the
    constituent-dimensioned shortmap loop.
  - New helper constituent_dim_file_var_names builds <base>_<name> in the
    as-registered / all-upper / all-lower spellings for constituent-dimensioned
    reads (e.g. cam_in_cflx_CO2 vs cam_in_cflx_co2).

M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_4D.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_bvd.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_cnst.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_constituent_dim.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_ddt.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_ddt2.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_ddt_array.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_host_var.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_initial_value.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_mf.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_no_horiz.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_noreq.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_param.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_protect.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_scalar.F90
M       test/unit/python/sample_files/write_init_files/phys_vars_init_check_simple.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_4D.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_bvd.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_cnst.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_constituent_dim.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_ddt.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_ddt2.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_ddt_array.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_host_var.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_initial_value.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_mf.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_no_horiz.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_noreq.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_param.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_protect.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_scalar.F90
M       test/unit/python/sample_files/write_init_files/physics_inputs_simple.F90
  - auto-generated file updates due to write_init_files.py update
```

derecho/intel/aux_sima:
  SMS_Ln9.ne3pg3_ne3pg3_mg37.FKESSLER.derecho_intel.cam-outfrq_se_cslam_multitape (Overall: NLFAIL) details:
    FAIL SMS_Ln9.ne3pg3_ne3pg3_mg37.FKESSLER.derecho_intel.cam-outfrq_se_cslam_multitape NLCOMP
  Pre-existing failure

derecho/gnu/aux_sima:
  SMS_Ln9.ne3pg3_ne3pg3_mg37.FADIAB.derecho_gnu.cam-outfrq_se_cslam (Overall: DIFF) details:
    FAIL SMS_Ln9.ne3pg3_ne3pg3_mg37.FADIAB.derecho_gnu.cam-outfrq_se_cslam BASELINE /glade/campaign/cesm/community/amwg/sima_baselines/latest_gnu: ERROR BFAIL some baseline files were missing
  Suspect test

derecho/nvhpc/aux_sima (test is run via Github workflow. Only run the test manually if we need to save new baselines): All PASS


src/data/write_init_files.py
```diff
@@ -340,10 +340,21 @@ def gather_ccpp_req_vars(cap_database, registry_constituents):
...
+                        # phys_var_stdnames row keyed to the scheme's local
+                        # name, which never matches a CAM snapshot field and
+                        # would pre-empt the fallback.
+                        pass
+                    else:
+                        # Constituent-related but not advected (e.g. tendency
+                        # variables): these are read as regular host variables,
+                        # not through the constituent loop.
                         in_vars[stdname] = cvar
                     # end if
```
peverwhee: how are non-advected (runtime) constituents handled?

peverwhee (commented):
One question.

Also: I'm hopeful that once we get rid of build-time constituents and everything is runtime, we'll be able to use the constituent diag_name to ease some of this logic.

src/data/write_init_files.py
jimmielin: Good question. They are registered in a case-sensitive way - I verified that `prescribed_aerosols`, ozone, etc. register as caps, but they get compared case-insensitive for the purpose of reading in ICs.

A tangential issue is that `ccpp_scheme_utils::ccpp_constituent_index` and our shim `ccpp_const_get_idx` in `to_be_ccppized` both compare case-sensitive, which is also a problem, so the full fix has to go into the framework side as well. I didn't do that because it'll break everything until we make atmos_phys changes to change everything to lowercase (the prescribed volcanic aerosols scheme queries `VOLC_MMR` all caps which would break...) & update the framework tag.

peverwhee: ah, ok. So yeah, many of our problems will go away when we remove build-time constituents.

In the interim, perhaps we modify our calls to ccpp_constituent_index to check both uppercase and lowercase?

jimmielin: Thanks! After discussing, we agreed to make the fix upstream in capgen to make the constituent index lookups case-insensitive, then the updated framework tag will be included in this PR.

peverwhee: opened PRs to remove more case sensitivity on the framework side (for both v0 and v1)
https://github.com/NCAR/ccpp-framework/pull/765
https://github.com/NCAR/ccpp-framework/pull/766

jimmielin: Hi @peverwhee, it crashes out of the box but not due to case sensitivity: it is because somewhere between the pinned tag and `develop` a change as added to handle dimension variables? I had to make https://github.com/ESCOMP/CAM-SIMA/pull/518/commits/5f84e58b663fcb21f4666122861f82dd62fb8257 to add `protected = True` to all dimension variables to make all the tests pass.

peverwhee: weird! not sure which commit would've done that.

peverwhee: we discussed this on monday, but, for posterity, @jimmielin tracked the `protected=true` change down to this commit in the framework: https://github.com/NCAR/ccpp-framework/commit/4c0539f28cd6a7cc254c993613931650fa8f301c

And we decided that we're fine with this fix as it is a net good.

peverwhee (approved):
thanks @jimmielin !

The new framework tag is available for you: sima_2026-07-15

src/data/write_init_files.py
```diff
@@ -340,10 +340,21 @@ def gather_ccpp_req_vars(cap_database, registry_constituents):
...
+                        # Advected constituents are allocated/registered by the
+                        # constituents object and read via the runtime
+                        # constituent loop (registry ic_file_input_names or
+                        # the cnst_/pbuf_ fallback).  Do NOT add them to the
+                        # host-variable list: that gives them a
+                        # phys_var_stdnames row keyed to the scheme's local
+                        # name, which never matches a CAM snapshot field and
+                        # would pre-empt the fallback.
+                        pass
+                    else:
```
nusbaume: This feels like a situation where the AI is over-explaining something (i.e. most of time a human wouldn't write a paragraph explaining the no-op path of a conditional).  Instead I might just change the conditional to its inverse and remove this `pass` section, especially since the `else` comment seems to explain it well enough.

```suggestion
                    if  not cvar.get_prop_value('advected'):           
```

Of course if you disagree just let me know, as I could certainly be convinced otherwise!

src/data/write_init_files.py
```diff
@@ -1180,26 +1192,31 @@ def write_phys_read_subroutine(outfile, host_dict, host_vars, host_imports,
...
-    # allows us to not enumerate all needed constituents from snapshots in the
-    # registry, yet allow reading their values from CAM snapshots.
-    outfile.write("call read_field(file, std_name, [character(len=std_name_len+5) :: std_name, 'cnst_'//trim(std_name), 'pbuf_'//trim(std_name)], 'lev', timestep, field_data_ptr(:,:,constituent_idx), mark_as_read=.false., error_on_not_found=.false., var_found=var_found)", 4)
+    outfile.comment("and cnst_, pbuf_ prefixes used by CAM snapshots (advected, non-advected) as input names.",4)
+    outfile.comment("Standard names are case-insensitive (capgen lowercases them) but netCDF names are not,",4)
+    outfile.comment("so also try the all-upper and all-lower case spellings of the constituent name:",4)
+    # The <std_name>, cnst_<std_name>, pbuf_<std_name> prefix default fallbacks --
+    # in the as-registered, all-upper, and all-lower case spellings -- allow us to
+    # not enumerate all needed constituents from snapshots in the registry, yet
+    # allow reading their values from CAM snapshots regardless of authored name case.
```
nusbaume: I might get rid of this comment block, as the text in the generated comments explains it well enough, at least in my opinion.

src/data/write_init_files.py
```diff
@@ -1492,25 +1510,30 @@ def write_phys_check_subroutine(outfile, host_dict, host_vars, host_imports,
...
-    # allows us to not enumerate all needed constituents from snapshots in the
-    # registry, yet allow reading their values from CAM snapshots.
-    outfile.write("call check_field(file, [character(len=std_name_len+5) :: std_name, 'cnst_'//trim(std_name), 'pbuf_'//trim(std_name)], 'lev', timestep, field_data_ptr(:,:,constituent_idx), std_name, min_difference, min_relative_value, is_first, diff_found)", 4)
+    outfile.comment("and cnst_, pbuf_ prefixes used by CAM snapshots (advected, non-advected) as input names.",4)
+    outfile.comment("Standard names are case-insensitive (capgen lowercases them) but netCDF names are not,",4)
+    outfile.comment("so also try the all-upper and all-lower case spellings of the constituent name:",4)
+    # The <std_name>, cnst_<std_name>, pbuf_<std_name> prefix default fallbacks --
+    # in the as-registered, all-upper, and all-lower case spellings -- allow us to
+    # not enumerate all needed constituents from snapshots in the registry, yet
+    # allow reading their values from CAM snapshots regardless of authored name case.
```
nusbaume: I would remove this comment-block as well.

src/physics/utils/physics_data.F90
```diff
@@ -661,6 +674,26 @@ subroutine read_constituent_dimensioned_field_2d(const_props, file, std_name, ba
...
+      ! case-sensitive, but a constituent name that is not enumerated in the
+      ! registry derives from its standard name, which is case-insensitive and
+      ! arrives lowercased from capgen (e.g. 'co2' for a field authored 'CO2').
+      ! Try the constituent name as registered plus its all-upper and all-lower
+      ! spellings; the base name keeps its case.
+      use string_utils, only: to_upper, to_lower
+
+      character(len=*), intent(in) :: base_name
+      character(len=*), intent(in) :: cname
+      character(len=256)           :: var_names(3)
```
nusbaume: Since this is SIMA-specific code we can use the shared CESM character lengths:

```suggestion
      use shr_kind_mod, only: cl=>shr_kind_cl
      ...
      character(len=cl)           :: var_names(3)
```

nusbaume (approved):
Thanks @jimmielin!  I have a few requests, mostly related to what looks like the AI being overly-verbose, but nothing that is strictly required or needs a re-review.

src/data/write_init_files.py
jimmielin: Thanks! I agree this is overly verbose but I think we could document the `if not` clause with a reason. Would this be better?

```python
                    # Do not add advected constituents to the host variable
                    # list so they are not included in phys_var_stdnames and
                    # can be read in the constituent path:
                    if not cvar.get_prop_value('advected'):
```

src/data/write_init_files.py
jimmielin: I also don't like this, so I am happy to remove. Would you think it's self-explanatory that some constituents in CAM are uppercase and some lowercase? I briefly considered changing this to something like
```
    # Try uppercase and lowercase variants for constituent names to match CAM
    # e.g., O3, NO2, NUMLIQ (most species) ... num_a1, num_a2 (aerosols)
```

src/data/write_init_files.py
jimmielin: Please see question above!

src/data/write_init_files.py
nusbaume: Yeah that comment block for the `if not` clause looks good to me, so feel free to add it.  Thanks!

src/data/write_init_files.py
nusbaume: I personally think it is self-explanatory, simply because having a system that tries the different case combinations is inherently more robust than a system that doesn't cover that space, regardless of what case(s) the constituents actually have.  

That being said if you don't feel the same then I am happy to include the comment you are proposing (as it is still way shorter than what was originally there).  Thanks!

src/physics/utils/physics_data.F90
jimmielin: Done!

PR #529, opened 2026-07-22
Add topography_statics_read host-side module to read topography dataset.

jimmielin:
Originator(s): @jimmielin 
  What: claude-fable:5
  How: wrote the initial version of the scheme.

- Add topography_statics_read to do a decomposed read (so it cannot be a CCPP scheme) of SGH, SGH30, LANDM_COSLAT
- Closes #534 

```
A       src/physics/utils/topography_statics_read.F90
  - new host-side read module
```

```
M       src/control/cam_comp.F90
  - call new topography_statics_read

M       src/data/registry.xml
  - move the topo fields to the top of the file away from "temporary initializations"
    although their standard names might still need some work (SGH, SGH30)

M       src/control/cam_initfiles.F90
  - closes #534 - check if topo file is not specified for aquap
```

derecho/intel/aux_sima:
```
  SMS_Ln9.ne3pg3_ne3pg3_mg37.FKESSLER.derecho_intel.cam-outfrq_se_cslam_multitape (Overall: NLFAIL) details:
    FAIL SMS_Ln9.ne3pg3_ne3pg3_mg37.FKESSLER.derecho_intel.cam-outfrq_se_cslam_multitape NLCOMP
  - pre-existing failure
```

derecho/gnu/aux_sima: All PASS

derecho/nvhpc/aux_sima (test is run via Github workflow. Only run the test manually if we need to save new baselines): All PASS


src/physics/utils/topography_statics_read.F90
```diff
@@ -0,0 +1,73 @@
...
+    logical                     :: found
+    character(len=*), parameter :: subname = 'topography_statics_read_file'
+
+    if (bnd_topo == unset_path_str) then
+      if (masterproc) then
+        write(iulog,*) trim(subname)//': no topo file (bnd_topo unset); '// &
+             'SGH, SGH30 and LANDM_COSLAT keep their initial values'
+      end if
+      return
+    end if
```
cacraigucar, line 37: Is this a legit way to run without having a topo file?  No one will see the warning in the huge amount of output that CAM spits out.

src/data/registry.xml
```diff
@@ -1436,6 +1427,32 @@
...
     </variable>
 
+    <!-- Topography dataset fields: default to 0 when no topo -->
+    <variable local_name="sgh"
+              standard_name="standard_deviation_of_subgrid_orography_for_orographic_gravity_wave_drag"
+              units="m" type="real" kind="kind_phys"
+              allocatable="allocatable">
+      <dimensions>horizontal_dimension</dimensions>
+      <initial_value>0.0_kind_phys</initial_value>
+      <ic_file_input_names>pbuf_SGH</ic_file_input_names>
```
cacraigucar, line 1437: This doesn't have plain SGH in the list, should it?

src/data/registry.xml
```diff
@@ -1436,6 +1427,32 @@
...
+      <initial_value>0.0_kind_phys</initial_value>
+      <ic_file_input_names>SGH30 pbuf_SGH30</ic_file_input_names>
+    </variable>
+    <variable local_name="landm"
+              standard_name="smoothed_land_area_fraction"
+              units="fraction" type="real" kind="kind_phys"
+              allocatable="allocatable">
+      <dimensions>horizontal_dimension</dimensions>
+      <initial_value>0.0_kind_phys</initial_value>
+      <ic_file_input_names>LANDM_COSLAT pbuf_LANDM</ic_file_input_names>
```
cacraigucar: Just confirming the LANDM_COSLAT name

cacraigucar (changes requested):

src/data/registry.xml
jimmielin: Thanks for confirming, yes it is mismatched with the pbuf name. https://github.com/ESCOMP/CAM/blob/5f582d41d2a109a5eb5369e0c8b7975d8e2ea4c4/src/physics/cam/physpkg.F90#L442-L443 It is `LANDM_COSLAT` from `infld` and the pbuf field is `LANDM`

**Edit** sorry, scratch that. The name is correct, but it shouldn't be in the registry, it is in the .F90 file so it is read from the topo file; this variable does not exist in the IC `ncdata` file. Removing it now.

src/physics/utils/topography_statics_read.F90
jimmielin: I think yes, at least for aquaplanet it is unset. 

CAM does not seem to abort https://github.com/ESCOMP/CAM/blob/5f582d41d2a109a5eb5369e0c8b7975d8e2ea4c4/src/physics/cam/physpkg.F90#L422; the if clause just goes through silently if not aquaplanet and topo is not set.

src/data/registry.xml
jimmielin: I don't think so, it's not a variable in the initial conditions file, only in the topo file (or read from `pbuf_SGH` when running from snapshot). But this does remind me that the `LANDM_COSLAT` shouldn't be there...

src/data/registry.xml
cacraigucar: Great - I know how I was generating the snapshot names, and it didn't seem to fit it.

src/physics/utils/topography_statics_read.F90
cacraigucar: Thanks for checking

cacraigucar (approved):

src/data/registry.xml
```diff
@@ -657,15 +657,6 @@
     </variable>
 
     <!-- temporaries from pbuf for snapshot usage. -->
```
peverwhee, line 659: now that you've removed landm, does this comment still apply to just the "ast" variable? maybe this is more a question for future us, when we eventually organize or split up the registry.

Side question: is this variable really "temporary" if we're using it for snapshot testing?

src/physics/utils/topography_statics_read.F90
peverwhee, line 37: following up on Cheryl's comment, it looks like CAM only uses the 0-value topo fields if the bnd_topo file is not unset AND the model is not running in aquaplanet mode (see snippet). Should we add that aquaplanet check here? Without it, could this potentially cause problems if someone ran an aquaplanet case with an F-compset user_nl_cam and erroneously specified a topography file?

```
if (associated(fh_topo) .and. .not. aqua_planet) then
...
else
      call pbuf_set_field(pbuf2d, sgh_idx, 0._r8)
      call pbuf_set_field(pbuf2d, sgh30_idx, 0._r8)
      call pbuf_set_field(pbuf2d, landm_idx, 0._r8)
end if
```

peverwhee (commented):
a couple questions for you, @jimmielin !

src/data/registry.xml
jimmielin: I suppose yes the comment applies, but not "temporary" in the sense that it should be removed. I agree we should get rid of the "temporary initializations for X" throughout `registry.xml` since if they're being used for snapshot testing they aren't really temporary?

src/physics/utils/topography_statics_read.F90
jimmielin: Thanks for the question! I am waffling a little on this. 

Looking at `namelist_definition_cam.xml` we already set `fh_topo` to `UNSET` when `aquaplanet="true"`:

https://github.com/ESCOMP/CAM-SIMA/blob/98ded70695ffc273d8a21bcf4f91346f177f9ba4/cime_config/namelist_definition_cam.xml#L193-L201

But in the scenario where `user_nl_cam` from a F-compset is erroneously used, CAM also doesn't guard against this properly. The SE dycore uses `topo` file `PHIS` unconditionally

https://github.com/ESCOMP/CAM-SIMA/blob/98ded70695ffc273d8a21bcf4f91346f177f9ba4/src/dynamics/se/dyn_comp.F90#L2114-L2120

so if we zeroed out the other vars here (to be consistent with CAM), the topo file is still being used *somehow* inside the dycore, so it's already half-consistent.

I wonder if the proper fix if we want to deny using topo file with aquap configurations is in `cam_initfiles_readnl` (where we read the topo file namelist) or when that runs initialization, to `endrun` when it happens.

If we really should add the check right here to be the same as CAM, personally I feel this line is really hard to read
```
if (associated(fh_topo) .and. .not. aqua_planet) then
```
so I am trying to simplify it to a behavior that is more easily parseable in my mind. 

Would you think this is better
```
if(aqua_planet .or. .not. associated(fh_topo)) then
  ! ... use zeros 
else 
  ! use topo file
end if
```
?

src/data/registry.xml
peverwhee: ok, yeah. i agree with that last part for sure. but no need to do any action on it now; we'll tackle it when we eventually (maybe) organize the registry...

src/physics/utils/topography_statics_read.F90
peverwhee: Thanks for the explanation / exploration!

I like the idea of adding a check in `cam_initfiles_readnl` to guard against improper aquaplanet/topo use. I'll open a wee issue.

peverwhee: https://github.com/ESCOMP/CAM-SIMA/issues/534

src/physics/utils/topography_statics_read.F90
```diff
@@ -0,0 +1,73 @@
+! Read the static subgrid topography fields consumed by physics (SGH, SGH30,
+! LANDM_COSLAT) from the topography dataset (namelist bnd_topo) into their
+! CAM-SIMA registry variables (sgh, sgh30, landm).
+!
+! Note: SE dycore handles PHIS itself (see set_phis)
```
peverwhee: one further question..

It looks like MPAS also handles PHIS (in the check_topography_data subroutine). Should this comment be changed to something like the following?
```suggestion
! Note: it is assumed that the dycore will handle PHIS (see set_phis in the SE dycore)
```

src/physics/utils/topography_statics_read.F90
jimmielin: Thanks Courtney! I just decided to implement it in https://github.com/jimmielin/CAM-SIMA/commit/cab96a1df4465a2836005d0203fb7de15411cc00 since it looks like a simple check, if you could take a look that would be great!

peverwhee: beautiful! thanks!

Now you can close an issue with this PR! How thrilling!

peverwhee (approved):
thanks @jimmielin !

PR #536, opened 2026-08-10
Remove hard-coded output units

peverwhee:
Originator(s): peverwhee
AI tools used (if applicable; please also add the "AI-generated code" label to the PR): n/a

1. Removes all hard-coded "6" and "*" output units; replace with "iulog" (iulog is being set to the intrinsic output unit rather than "6")
2. Applies fixes for issues raised by Fortitude on the files modified.

Describe any changes made to build system: n/a

List all files added and what they do: n/a

```
M   src/cpl/nuopc/atm_comp_nuopc.F90
M   src/dynamics/se/dycore/comp_ctr_vol_around_gll_pts.F90
M   src/dynamics/se/dycore/edge_mod.F90
M   src/dynamics/se/dycore/element_mod.F90
M   src/dynamics/se/dycore/fvm_mod.F90
M   src/dynamics/se/dycore/schedule_mod.F90
M   src/dynamics/se/dycore/vertremap_mod.F90
- remove "6"
- apply fixes for Fortitude linting errors

M   src/dynamics/se/dycore/bndry_mod.F90
M   src/dynamics/se/dycore/coordinate_systems_mod.F90
M   src/dynamics/se/dycore/fvm_overlap_mod.F90
M   src/dynamics/se/dycore/fvm_reconstruction_mod.F90
M   src/dynamics/se/dycore/hybrid_mod.F90
M   src/dynamics/se/dycore/interpolate_mod.F90
- remove "*"
- apply fixes for Fortitude linting errors

M  test/fortitude.toml
- ignore rule about trailing decimals

```

derecho/intel/aux_sima:

derecho/gnu/aux_sima:


jimmielin (approved):
Thanks @peverwhee !

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -189,16 +189,17 @@ end subroutine SetServices
...
   !===============================================================================
   subroutine InitializeP0(gcomp, importState, exportState, clock, rc)
-    type(ESMF_GridComp)   :: gcomp
-    type(ESMF_State)      :: importState, exportState
-    type(ESMF_Clock)      :: clock
-    integer, intent(out)  :: rc
+    type(ESMF_GridComp), intent(inout) :: gcomp
+    type(ESMF_State),    intent(inout) :: importState
+    type(ESMF_State),    intent(inout) :: exportState
+    type(ESMF_Clock),    intent(inout) :: clock
```
cacraigucar: I know this isn't your code changes, but as I was checking the intent, it appears that these variables are not used.  Perhaps this should be cleaned up (remove the variables not used)?  I'm not sure how far out in the code this would extend.  @nusbaume and @jimmielin - what are your thoughts?  Changes that are decided here may need to be applied to other routines as well in this module.

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -330,10 +332,10 @@ subroutine InitializeRealize(gcomp, importState, exportState, clock, rc)
     use physics_grid, only : lsize => columns_on_task
 
     ! input/output variables
-    type(ESMF_GridComp)  :: gcomp
-    type(ESMF_State)     :: importState
-    type(ESMF_State)     :: exportState
-    type(ESMF_Clock)     :: clock
+    type(ESMF_GridComp), intent(inout) :: gcomp
+    type(ESMF_State),    intent(inout) :: importState
```
cacraigucar: If I were to guess, I'd think that importState would be an intent (in) variable.  Did you check this when you made this change?

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -1065,11 +1068,12 @@ subroutine ModelAdvance(gcomp, rc)
     logical                 :: nlend       ! Flag signaling last time-step
     integer                 :: lbnum
     integer                 :: localPet, localPeCount
-    logical                 :: first_time = .true.
+    logical                 :: first_time
     logical                 :: do_ncdata_check  !Flag notifying SIMA if it is OK to perform a snapshot check
     character(len=*),parameter  :: subname=trim(modName)//':(ModelAdvance) '
     !-------------------------------------------------------------------------------
 
+    first_time = .true.
```
cacraigucar: I don't think this works the way you want.  In the old code, first_time was a module variable which remembered being set to false.  Now it is a local variable that is being set to true everytime it is called.

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -2112,7 +2119,7 @@ end subroutine cam_write_clockrest
   subroutine cam_set_mesh_for_single_column(scol_lon, scol_lat, mesh, rc)
 
     ! Generate a mesh for single column
-    use netcdf
+!    use netcdf
```
cacraigucar: Should this just be deleted instead of being commented out?

src/dynamics/se/dycore/comp_ctr_vol_around_gll_pts.F90
```diff
@@ -971,7 +969,7 @@ subroutine construct_cv_duel(elem,hybrid,nets,nete)
           vert(:)%x = 0
           vert(:)%y = 0
           vert(:)%z = 0
-          if (i.eq.0.and.j.eq.0) then
+          if (i==0.and.j==0) then
```
cacraigucar: I"m not sure if this is a fortitude rule or not, but I prefer a space before and after `.and.`)  If this is a fortitude rule, I'd like to propose changing it as the spaces break each conditional into an easy to see section

src/dynamics/se/dycore/comp_ctr_vol_around_gll_pts.F90
```diff
@@ -943,7 +941,7 @@ subroutine construct_cv_duel(elem,hybrid,nets,nete)
           vert(5) = cv(i-1, j+1)
           vert(6) = cv(i-1, j  )
           p = j
-          if (p.eq.0) p=1
+          if (p==0) p=1
```
cacraigucar: Could we please put spaces around `==`?  This would be for all the newly added places below

src/dynamics/se/dycore/comp_ctr_vol_around_gll_pts.F90
```diff
@@ -980,7 +978,7 @@ subroutine construct_cv_duel(elem,hybrid,nets,nete)
             vert(m+5) = cv(i-1, j+1)  !
             vert(m+6) = cv(i-1, j  )  !     X       X     (+1,-1)  1
             m = m + 6
-            if (mlt(swest).ne.0) then
+            if (mlt(swest)/=0) then
```
cacraigucar: Could we please put spaces around `/=` here and below?

src/dynamics/se/dycore/edge_mod.F90
```diff
@@ -399,61 +399,61 @@ subroutine initEdgeBuffer_r8(par,edge,elem,nlyr, bndry_type,nthreads,CardinalLen
...
-       endif
-    enddo
+       end if
+    end do
 
     icInter=0
     icIntra=0
     do icycle=1,nSendCycles
-       if(pSchedule%SendCycle(icycle)%onNode .eqv. .FALSE.) then
+       if(pSchedule%SendCycle(icycle)%onNode .eqv. .false.) then
```
cacraigucar: Since we are updating our conditional checks, we should perhaps use `if (.not. pSchedule%SendCycle(icycle)%onNode) then`.  I believe `.eqv.` is not preferred for comparing to `.true.` or `.false.`

src/dynamics/se/dycore/edge_mod.F90
```diff
@@ -1301,11 +1301,10 @@ subroutine edgeVunpackVert(edge,v,ielem)
             v(1,1,1)%y=edge%receive(nce+isw+1)
             v(1,1,1)%z=edge%receive(2*nce+isw+1)
             exit
-        else
-            v(1,1,1)%x=0_r8
-            v(1,1,1)%y=0_r8
-            v(1,1,1)%z=0_r8
-        endif
+         end if
```
cacraigucar: I don't think this should have been replaced with an `end if`, but should have remained `else` as the next 3 lines overwrite what was just written

src/dynamics/se/dycore/edge_mod.F90
```diff
@@ -1317,11 +1316,10 @@ subroutine edgeVunpackVert(edge,v,ielem)
             v(2,np,1)%y=edge%receive(nce+ise+1)
             v(2,np,1)%z=edge%receive(2*nce+ise+1)
             exit
-        else
-            v(2,np,1)%x=0_r8
-            v(2,np,1)%y=0_r8
-            v(2,np,1)%z=0_r8
-        endif
+        end if
```
cacraigucar, line 1322: Same bug here - keep the else?

src/dynamics/se/dycore/edge_mod.F90
```diff
@@ -1333,11 +1331,10 @@ subroutine edgeVunpackVert(edge,v,ielem)
             v(3,np,np)%y=edge%receive(nce+ine+1)
             v(3,np,np)%z=edge%receive(2*nce+ine+1)
             exit
-        else
-            v(3,np,np)%x=0_r8
-            v(3,np,np)%y=0_r8
-            v(3,np,np)%z=0_r8
-        endif
+        end if
```
cacraigucar, line 1337: Same bug here, keep the else?

src/dynamics/se/dycore/edge_mod.F90
```diff
@@ -1349,11 +1346,10 @@ subroutine edgeVunpackVert(edge,v,ielem)
             v(4,1,np)%y=edge%receive(nce+inw+1)
             v(4,1,np)%z=edge%receive(2*nce+inw+1)
             exit
-        else
-            v(4,1,np)%x=0_r8
-            v(4,1,np)%y=0_r8
-            v(4,1,np)%z=0_r8
-        endif
+        end if
```
cacraigucar, line 1352: Same bug here - keep the else?

cacraigucar (changes requested):
Partial review as I want to get onto PUMAS.  I still need to review fvm_mod.F90, schedule_mod.F90 and vertremap_mod.F90.  A number of cosmetic changes and I believe a few actual bugs

src/cpl/nuopc/atm_comp_nuopc.F90
peverwhee: i did not know about the save thing! i'll change it back.

src/dynamics/se/dycore/edge_mod.F90
peverwhee: i think the "end if"s are ok because the "if" block has an exit so you'd never keep going after executing those lines. fortitude flagged it as unnecessary.

src/cpl/nuopc/atm_comp_nuopc.F90
peverwhee: Yeah, I think that the interfaces for each of these is standard across all the components. I picked "inout" for all of them for consistency and to avoid future problems if we ever need to do things with these unused variables.

nusbaume: Yeah, I think trying to get rid of them would require us to make changes in CMEPS itself, which is probably too much effort at this stage.  However, given that we aren't using them at all, would `intent(in)` be the safest choice for now? I would at least do that for the `clock` variable as I doubt SIMA would ever modify that.

src/dynamics/se/dycore/edge_mod.F90
cacraigucar: I should know better than to question a tool.  I missed the exit

src/cpl/nuopc/atm_comp_nuopc.F90
peverwhee: ok i changed all the of the unused ones to intent(in) (as well as clock)

src/cpl/nuopc/atm_comp_nuopc.F90
peverwhee: removed; "save" keyword added to variable declaration

src/cpl/nuopc/atm_comp_nuopc.F90
peverwhee: removed!

src/dynamics/se/dycore/comp_ctr_vol_around_gll_pts.F90
peverwhee: spacing added in this file.

src/dynamics/se/dycore/edge_mod.F90
peverwhee: good call! changed.

src/dynamics/se/dycore/edge_mod.F90
peverwhee: this did highlight to me that the indentation is off here - fixed!

cacraigucar: I'm wondering if it makes sense to initialize the values to zero once at the beginning of the loop instead of on every iteration?  Unless, once again I am missing something subtle on why it needs to happen within the loop.

peverwhee: good point! changed!

Thanks for pointing out such a useful cleanup! I genuinely feel better about this whole exercise now.

jimmielin (changes requested):
I would prefer that the `fortitude-errors.txt` and the behavior change in a separate PR. Right now it appears that the behavior change being bundled into a PR that otherwise fixes Fortitude issues.

src/dynamics/se/dycore/vertremap_mod.F90
```diff
@@ -294,17 +301,17 @@ subroutine remap1_nofilter(Qdp,nx,qsize,dp1,dp2)
         zv1 = 0
         do k=1,nlev
           if (zgam(k+1)>1_r8) then
-            WRITE(*,*) 'r not in [0:1]', zgam(k+1)
+            write(*,*) 'r not in [0:1]', zgam(k+1)
```
cacraigucar: This write should use iulog

src/dynamics/se/dycore/comp_ctr_vol_around_gll_pts.F90
```diff
@@ -392,7 +391,7 @@ subroutine gll_grid_write(elem, grid_format, filename_in)
 #endif
 !!XXgoldyXX: ^ debug only
```
cacraigucar, line 392: I just saw this - should we be trying to remove the XXgoldyXX comments as we encounter them?  They were reminders Steve put in for himself to come back and do something with.

src/cpl/nuopc/atm_comp_nuopc.F90
cacraigucar: I still don't think this is working the way it used to (unless github is not displaying your changes properly).  In the old way of doing things, the module level variable was initialized to true.  Then internally in the lower down subroutine, the variable was set to false.  Module level variables persist even when the module is exited.  So once it is set to false, it remains false.  Now it appears that you are setting it to true every time you enter this subroutine which is not what is wanted.  I believe that initialization of a module level variable is a good thing to be able to do and if Fortitude is flagging it, I'd be curious as to why.

cacraigucar (changes requested):
Review completed with a couple of minor points and continued discussion on the initialization of module level variable (which I think is a legit thing to do)

jimmielin (approved):
Thanks @peverwhee! Sorry for nitpicking about this being in a separate PR I just feel it deserves some further discussion and not block these changes!

src/dynamics/se/dycore/vertremap_mod.F90
peverwhee: good point! fixed this and others throughout the SE dycore code

src/dynamics/se/dycore/comp_ctr_vol_around_gll_pts.F90
peverwhee: If we remove these, I'd say we should capture whatever Steve was trying to do with PIO3D in an issue. @nusbaume - any ideas?

nusbaume: Sadly it looks like these `XXgoldyXX ` comments exist in CAM itself, so they likely predate any CAM-SIMA work.  Maybe @PeterHjortLauritzen knows?

As an aside the `PIO3D` CPP def is always defined, so the "debug" code is being run if the Fortran subroutines are ever called.

src/cpl/nuopc/atm_comp_nuopc.F90
peverwhee: ok just kidding! Declaring these intents turned out to be a minefield because the arguments need to match the subroutine definitions in NUOPC in order to work.

I just backed all of the intent changes out...

cacraigucar (approved):

src/dynamics/se/dycore/edge_mod.F90
```diff
@@ -2464,12 +2455,12 @@ subroutine ghostVunpack3d(g, v, vlyr, kptr, desc, sw, se, nw, ne, mult)
     real (kind=r8), intent(out)    :: se(   g%np : g%np+g%nhc, 1-g%nhc : 1,          vlyr, max_corner_elem-1)
     real (kind=r8), intent(out)    :: ne(   g%np : g%np+g%nhc,    g%np : g%np+g%nhc, vlyr, max_corner_elem-1)
     real (kind=r8), intent(out)    :: nw(1-g%nhc : 1,             g%np : g%np+g%nhc, vlyr, max_corner_elem-1)
-    type (EdgeDescriptor_t)               :: desc
+    type (EdgeDescriptor_t), intent(inout) :: desc
```
jimmielin: I think this should be `intent(in)` since `desc%` is not modified at all in here. 

If `desc` is changed to `intent(in)` here all of its callers upstream do not need to be widened to `inout`, e.g., `elem` in `comp_ctr_vol_around_gll_pts.F90` then up the chain, `InitControlVolumes_duel`, ...

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -403,12 +405,13 @@ subroutine InitializeRealize(gcomp, importState, exportState, clock, rc)
     logical                 :: branch_run                        ! branch from a previous run; requires a restart file
     character(len=CL)       :: tempc1,tempc2
     integer                 :: shrlogunit          ! original log unit
-    real(r8)        , parameter :: radtodeg = 180.0_r8/shr_const_pi
+    real(r8)        , parameter :: radtodeg = 180_r8/shr_const_pi
```
jimmielin:
```suggestion
    real(r8)        , parameter :: radtodeg = 180.0_r8/shr_const_pi
```

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -743,8 +746,8 @@ subroutine InitializeRealize(gcomp, importState, exportState, clock, rc)
           end do
 
           ! obtain internally generated cam lats and lons
-          lon(:) = 0._r8
-          lat(:) = 0._r8
+          lon(:) = 0_r8
+          lat(:) = 0_r8
```
jimmielin: `0._r8`

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -756,13 +759,13 @@ subroutine InitializeRealize(gcomp, importState, exportState, clock, rc)
           ! error check differences between internally generated lons and those read in
           do n = 1,lsize
              if (abs(lonMesh(n) - lon(n)) > grid_tol .and. .not. &
-                  abs(abs(lonMesh(n) - lon(n))- 360._r8) < grid_tol) then
-                write(6,100)n,lon(n),lonMesh(n), abs(lonMesh(n)-lon(n))
+                  abs(abs(lonMesh(n) - lon(n))- 360_r8) < grid_tol) then
```
jimmielin: `360.0_r8`

src/cpl/nuopc/atm_comp_nuopc.F90
```diff
@@ -1787,7 +1791,7 @@ subroutine cam_read_srfrest( gcomp, clock, rc )
                       write(iulog,*)'cam_read_srfrest warning: field ',trim(varname),' is not on restart file'
                       write(iulog,*)'for backwards compatibility will set it to 0'
                    end if
-                   tmpptr(:) = 0._r8
+                   tmpptr(:) = 0_r8
```
jimmielin: `0.0_r8`

jimmielin (approved):
Thanks @peverwhee! Just flagging the integer literals as we discussed in chat. 

For future reference, F2003 4.4.1 states
> The optional kind type parameter following digit-string specifies the kind type parameter of the integer constant; if it is not present, the constant is of type default integer.

More rabbit holes about integer literals in our chat.

cacraigucar (approved):
Reported in person the last few items.   All _r8 appear to now have `.` in them along with the one where the _r8 was accidentally deleted completely.

src/dynamics/se/dycore/edge_mod.F90
peverwhee: changed here and in comp_ctr_vol_around_all_pts.F90

